PR #1, opened 2016-02-29
Add EventHub templates.

MikeStall:


Bindings/bindings.json
```diff
@@ -7,23 +7,67 @@
         "triggerRefHelp": "You can use brackets ({value}) and percent symbols (%value%) to format your path based on parameters from the trigger object."
     },
     "bindings": [
-        {
-            "type": "timerTrigger",
-            "displayName": "Timer",
-            "direction": "trigger",
+      {
```
davidebbo: Spacing is inconsistent with rest of file. I'm fine changing to 2 spaces, but then let's convert whole file

MikeStall: I'm just using the VS defaults.

davidebbo: Strange, VS default was 4 for me. But in any case, let's not fix in same file :)

davidebbo:
:shipit: Just the spacing comment. That can be done separately

PR #2, opened 2016-03-02
Function binding metadata format changes, and also introducing sln file

mathewc:
Absorbing the function binding metadata file changes


ahmelsayed:
why a sln file? 
Also I'm just confirming my understanding, trigger vs input is only based on suffix `Trigger` right?

mathewc:
Sln file because we need some way to view/manage these files. In my case, I needed to do a bunch of find/replace modificaitons for this change, and the project structure greatly facilitated things. What would you prefer?

ahmelsayed:
oh, I see. I usually just use sublime text, but that's fine too if you wanna use VS.

mathewc:
All bindings have a direction in/out or inout. A trigger is always an 'in' binding. We might even default the direction allowing it not to be specified for triggers, but I've been explicit.

ahmelsayed:
:shipit:

PR #15, opened 2016-03-09
Updated core timer and http templates

soninaren:


paulbatum:
Looks good to me.

FYI @davidebbo this resolves the HTTP trigger issue we were seeing. Specifically, the addition of a "none" option for the webhook type.

christopheranderson:
Looks good. I can merge when ready.

PR #16, opened 2016-03-09
Updated blob trigger for C# and Node

soninaren:


Templates/BlobTrigger-CSharp/sample.dat
```diff
@@ -1,3 +1 @@
-{
-    "id":"foobar123"
-}
\ No newline at end of file
+"foobar123"
```
fabiocav: Can we change this to something like "My sample data"

soninaren:
Updated the default names for parameters

fabiocav:
:+1:

PR #17, opened 2016-03-09
Applying label changes and introducing labels for bound parameters

mattchenderson:


christopheranderson:
Copy edits + 1 new setting for better portal labels. Nothing breaking here. :+1:

Bindings/bindings.json
```diff
@@ -12,21 +12,22 @@
...
+                    "help": "A cron expression is a way of expressing a schedule. Please consult the documentation to learn more."
                 },
                 {
                     "name": "runOnStartup",
                     "value": "boolean",
                     "defaultValue": false,
-                    "label": "Run on start up?",
-                    "help": "When true, your function will run on start. Useful for debugging."
+                    "label": "Should this function run on start up?",
+                    "help": "When true, your function will run on startup. This may be useful for debugging."
```
mathewc: Hmmm, we should think more about the notion of "startup". While it makes sense in core SDK continuous/always on scenarios, I don't think "startup" is part of the conceptual model for Functions. Also not sure about the note about debugging.

Bindings/bindings.json
```diff
@@ -225,8 +232,8 @@
                             "display": "Generic"
                         }
                     ],
-                    "label": "Is this a WebHook? Select which type of WebHook",
-                    "help": "If your type isn't listed here, you can leave this blank."
+                    "label": "Is this a WebHook? If so, what type of webhook is it?",
```
mathewc: Fix "webhook" casing. Make consistent.

Bindings/bindings.json
```diff
@@ -330,21 +339,22 @@
             "displayName": "Azure Storage Table",
             "direction": "in",
             "defaultParameterName": "myTable",
+            "parameterNamePrompt":"Name the row collection for use in your code",
```
mathewc: The binding also supports singleton (specifying rowkey as well).

PR #20, opened 2016-03-10
Update JavaScript Samples to use new format & their template metadata

christopheranderson:
Updates:
 - Changed the format to use new "context first" format (https://github.com/Azure/azure-webjobs-sdk-script/pull/70)
 - Removed output bindings and cleaned up function.json
 - Modified metadata.json to use current format

### DIFF
--- a/Templates/EventHubTrigger/function.json
+++ b/Templates/EventHubTrigger/function.json
@@ -5,12 +5,6 @@
       "direction": "in",
       "path": "samples-workitems",
       "connectionString": "<your EventHub receiving connection string here>"
-    },
-    {
-      "type": "blob",
-      "name": "receipt",
-      "direction": "out",
-      "path": "samples-workitems/{id}"
     }
   ]
 }
--- a/Templates/EventHubTrigger/function.json
+++ b/Templates/EventHubTrigger/function.json
@@ -2,6 +2,7 @@
   "bindings": [
     {
       "type": "eventHubTrigger",
+      "name":"myEventHubTrigger",
       "direction": "in",
       "path": "samples-workitems",
       "connectionString": "<your EventHub receiving connection string here>"
--- a/Templates/EventHubTrigger/index.js
+++ b/Templates/EventHubTrigger/index.js
@@ -1,5 +1,5 @@
-﻿module.exports = function (context, workItem) {
-    context.log('Node.js eventhub trigger function processed work item ' + workItem);
+﻿module.exports = function (context, myEventHubTrigger) {
+    context.log('Node.js eventhub trigger function processed work item ' + myEventHubTrigger);
 
     context.done();
 }
--- a/Templates/EventHubTrigger/index.js
+++ b/Templates/EventHubTrigger/index.js
@@ -1,5 +1,5 @@
-﻿module.exports = function (workItem, context) {
+﻿module.exports = function (context, workItem) {
     context.log('Node.js eventhub trigger function processed work item ' + workItem);
 
-    context.done(null, workItem);
+    context.done();
 }
--- a/Templates/HttpTrigger/function.json
+++ b/Templates/HttpTrigger/function.json
@@ -2,11 +2,13 @@
   "bindings": [
     {
       "type": "httpTrigger",
-      "direction": "in"
+      "direction": "in",
+      "name": "req"
     },
     {
       "type": "http",
-      "direction": "out"
+      "direction": "out",
+      "name": "res"
     }
   ]
 }
--- a/Templates/HttpTrigger/index.js
+++ b/Templates/HttpTrigger/index.js
@@ -1,5 +1,5 @@
 ﻿module.exports = function (context, req) {
-    context.log('Node.js HTTP trigger function processed a request. Name=' + req.query.name);
+    context.log('Node.js HTTP trigger function processed a request. Name=' + req.originalUrl);
 
     if (typeof req.query.name == 'undefined') {
         context.res = {
--- a/Templates/HttpTrigger/index.js
+++ b/Templates/HttpTrigger/index.js
@@ -1,10 +1,17 @@
-﻿module.exports = function (req, context) {
+﻿module.exports = function (context, req) {
     context.log('Node.js HTTP trigger function processed a request. Name=' + req.query.name);
 
     if (typeof req.query.name == 'undefined') {
-        context.done(null, "Please pass a name on the query string");
+        context.res = {
+            status: 400,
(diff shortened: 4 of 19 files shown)


Templates/HttpTrigger/function.json
```diff
@@ -2,11 +2,13 @@
   "bindings": [
     {
       "type": "httpTrigger",
-      "direction": "in"
+      "direction": "in",
+      "name": "req"
```
davidebbo: Eventually, we should not need to have those explicit names.

Templates/ManualTrigger/index.js
```diff
@@ -1,4 +1,4 @@
-﻿module.exports = function (input, context) {
-    context.log('Node.js manually triggered function called with input ' + input);
+﻿module.exports = function (context, input) {
+    context.log('Node.js manually triggered function called with input ' + JSON.stringify(input));
```
davidebbo: Why do we now need explicit stringify?

Templates/HttpTrigger/function.json
christopheranderson: We do already default those names, so I could remove this. However, when you add a new binding, we currently always include the name, thus I though it was a good idea our examples should always include the name, as well, even though we have good defaults.

Templates/WebHook-GitHub/index.js
```diff
@@ -1,4 +1,8 @@
 ﻿module.exports = function (body, context) {
     context.log('GitHub WebHook triggered! ' + body.comment.body);
-    context.done(null, 'New GitHub comment: ' + body.comment.body);
+    context.done(null, {
+        res: {
+            body: 'New GitHub comment: ' + JSON.stringify(body.comment.body, null, " ")
```
davidebbo: That feels really ugly. But we know we need to improve this. I want `context.done(null, body.comment.body);`

Templates/HttpTrigger/function.json
davidebbo: Understood.

Templates/ManualTrigger/index.js
christopheranderson: You don't need to if it is a string. However, if your input object is a JSON, it will log "[Object object]". This has always been the case; our samples just always assumed string for input data type. I figured we should handle both strings and JSON.

davidebbo:
:shipit:

Templates/WebHook-GitHub/index.js
christopheranderson: Yes, this is technically correct, but quite heavy. Do you prefer the `context.res.body = "text"` format for examples? It avoid the bulky property bag.

Templates/EventHubTrigger/function.json
```diff
@@ -5,12 +5,6 @@
       "direction": "in",
```
soninaren: name is missing

Templates/EventHubTrigger/index.js
```diff
@@ -1,5 +1,5 @@
-﻿module.exports = function (workItem, context) {
+﻿module.exports = function (context, workItem) {
```
soninaren: name in the function.json should replace workItem

Templates/WebHook-GitHub/index.js
davidebbo: It's a bit cleaner, but still only transitional. Once we assume Hook output is JSON, it naturally becomes much simpler.

Templates/EventHubTrigger/function.json
christopheranderson: :+1: addressing this now

Templates/TimerTrigger/function.json
```diff
@@ -2,7 +2,7 @@
   "disabled": false,
   "bindings": [
     {
-      "name": "myTimer",
+      "name": "timerinfo",
```
fabiocav: The C# templates all seem to use the "my*" naming convention. We should try to be consistent with those.

christopheranderson: :+1: Agreed, I was trying to do the wrong thing here. I'll send a change.

Templates/HttpTrigger/index.js
```diff
@@ -1,10 +1,17 @@
-﻿module.exports = function (req, context) {
+﻿module.exports = function (context, req) {
     context.log('Node.js HTTP trigger function processed a request. Name=' + req.query.name);
```
soninaren: We should change this to URI, because if the query string is not present then user will see "undefined" the logs

christopheranderson: This is what C# does, as well, so I'll agree with that change. :+1:

Templates/TimerTrigger/function.json
soninaren: this should match the defaultParameterName in binding.json
 "defaultParameterName": "myTimer",

Templates/TimerTrigger/index.js
```diff
@@ -1,5 +1,11 @@
-﻿module.exports = function (context) {
+﻿module.exports = function (context, timerinfo) {
     var timeStamp = new Date().toISOString();
+
```
soninaren: same here 
"defaultParameterName": "myTimer",

Templates/TimerTrigger/function.json
christopheranderson: I've updated this in https://github.com/christopheranderson/azure-webjobs-sdk-templates/commit/b732e497d2fe624ee2dfe0c96b3f587c9f5ff03b

Templates/TimerTrigger/index.js
christopheranderson: :+1: Updated here: https://github.com/christopheranderson/azure-webjobs-sdk-templates/commit/b732e497d2fe624ee2dfe0c96b3f587c9f5ff03b

Templates/WebHook-Generic/index.js
```diff
@@ -1,4 +1,6 @@
-﻿module.exports = function (body, context) {
+﻿module.exports = function (context, body) {
     context.log('Webhook was triggered!');
```
soninaren: not sure if we need to change body to req

christopheranderson: WebHooks are special cased to provide the body as the input parameter, so this is an exception to that rule.

soninaren: :+1:

christopheranderson:
@soninaren + everyone online seems cool with the changes. Merging.

PR #21, opened 2016-03-10
adding allow list for new core template

mattchenderson:


mattchenderson:
This makes the C# queuetrigger template show the right fields during create-time.

christopheranderson:
@soninaren & @alrod Please give this a look

soninaren:
Looks good to me

PR #26, opened 2016-03-10
Changing log messages from "CSharp" to "C#"

fabiocav:


fabiocav:
Changed from string.Format to string interpolation. Also fixed timer string format.

soninaren:
Looks good merging it.

PR #30, opened 2016-03-12
Add C# WebHook template

davidebbo:
That's the cleanest I could do with current runtime. Later, I'd like to find a way to directly receive the dynamic `data` as a function param instead of having to manually use json.net.

### DIFF
--- a/Templates/WebHook-Generic-CSharp/run.csx
+++ b/Templates/WebHook-Generic-CSharp/run.csx
@@ -1,0 +1,25 @@
+﻿#r "Newtonsoft.Json"
+
+using System;
+using System.Net;
+using System.Net.Http;
+using Microsoft.Azure.WebJobs.Host;
+using Newtonsoft.Json;
+
+public static object Run(HttpRequestMessage req, TraceWriter log)
+{
+    log.Verbose($"Webhook was triggered!");
+
+    string jsonContent = req.Content.ReadAsStringAsync().Result;
+    dynamic data = JsonConvert.DeserializeObject(jsonContent);
+
+    if (data.first == null || data.last == null) {
+        return req.CreateResponse(HttpStatusCode.BadRequest, new {
+            error = "Please pass first/last properties in the input object"
+        });
+    }
+
+    return req.CreateResponse(HttpStatusCode.OK, new {
+        greeting = $"Hello {data.first} {data.last}!"
+    });
+}
--- a/Templates/WebHook-Generic-CSharp/run.csx
+++ b/Templates/WebHook-Generic-CSharp/run.csx
@@ -3,14 +3,15 @@
 using System;
 using System.Net;
 using System.Net.Http;
+using System.Threading.Tasks;
 using Microsoft.Azure.WebJobs.Host;
 using Newtonsoft.Json;
 
-public static object Run(HttpRequestMessage req, TraceWriter log)
+public static async Task<object> Run(HttpRequestMessage req, TraceWriter log)
 {
     log.Verbose($"Webhook was triggered!");
 
-    string jsonContent = req.Content.ReadAsStringAsync().Result;
+    string jsonContent = await req.Content.ReadAsStringAsync();
     dynamic data = JsonConvert.DeserializeObject(jsonContent);
 
     if (data.first == null || data.last == null) {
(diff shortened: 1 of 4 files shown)


Templates/WebHook-Generic-CSharp/run.csx
```diff
@@ -0,0 +1,25 @@
...
+using System.Net;
+using System.Net.Http;
+using Microsoft.Azure.WebJobs.Host;
+using Newtonsoft.Json;
+
+public static object Run(HttpRequestMessage req, TraceWriter log)
+{
+    log.Verbose($"Webhook was triggered!");
+
+    string jsonContent = req.Content.ReadAsStringAsync().Result;
```
mathewc: I think you can do something like `dynamic data = await response.Content.ReadAsAsync<object>();`. Have you tried that?

mathewc: Also, you're calling an Async function - don't you have to await that?

mathewc:
Looks fine to me, if the dynamic ReadAsAsync<object> I suggested doesn't work.

Templates/WebHook-Generic-CSharp/run.csx
davidebbo: Strangely, I can't make it work in the context of my function.

    error CS1061: 'HttpContent' does not contain a definition for 'ReadAsAsync' and no extension method 'ReadAsAsync' accepting a first argument of type 'HttpContent' could be found (are you missing a using directive or an assembly reference?)

davidebbo: Oh, seems we're missing a reference to `System.Net.Http.Formatting.dll`. @fabiocav should we add this?

fabiocav: ReasAsAsync is an extension method defined in System.Net.Http.Formatting. You'd need to add that reference as well :(

fabiocav: Commented at the same time :)

Probably. Those extension methods will are very handy and even more important in this context.

davidebbo: Note that `#r "System.Net.Http.Formatting"` does not work, probably because it's not in your list. We should just add it by default. Opened https://github.com/Azure/azure-webjobs-sdk-script/issues/91.

mathewc: My wife just yelled at me to get off the computer. Have fun guys. Bye.

davidebbo:
So there are as many as 3 distinct issues with `req.Content.ReadAsAsync<object>()`:

1, Missing assembly ref to `System.Net.Http.Formatting`
2. The run button doesn't send right json content type, causing `No MediaTypeFormatter is available to read an object of type 'Object' from content with media type 'plain/text'`. Opened https://github.com/projectkudu/WebJobsPortal/issues/67
3. The kind of object it returns doesn't allow for checking `if (data.first == null)` when first doesn't exist. Fails with `Cannot perform runtime binding on a null reference`.

The first two are fixable, but the 3rd is tricky. The beauty of the json.net object is that it's more flexible. Not all dynamic are equal!

davidebbo:
I'll just stick with my original till we figure this out. It runs pretty nicely!

Templates/WebHook-Generic-CSharp/run.csx
fabiocav: Yeah, the assembly is not a framework assembly, so it doesn't get automatically resolved by name with a #r unless we add it to the shared assemblies we've introduced. The other option is to add it automatically (as we do with System.Web.Http). Thoughts?

davidebbo: Yes, I think we should. We already have it loaded in the runtime, so I doubt it will affect compilation perf much.

fabiocav:
An ExpandoObject should work well ReasAsAsync.
I'll submit a fix to add the `System.Net.Http.Formatting` assembly.

davidebbo:
@fabiocav ExpandoObject works pretty poorly. e.g. you have to do [crazy things](http://stackoverflow.com/questions/2839598/how-to-detect-if-a-property-exists-on-an-expandoobject) to check property existence:

```c#
var expandoObject = ...;
if(((IDictionary<String, object>)expandoObject).ContainsKey("SomeMember")) {
    // expandoObject.SomeMember exists.
}
```

But the object returns by Json.NET works really well here. :)

fabiocav:
Yeah, I removed my suggestion shortly after I commented once I thought more about it :)

You're right, using Json.NET in this case works well and it's a good thing to have samples on.

Issue #31, opened 2016-03-13
Fix Node.js HTTP template log output


mathewc:
The Name= output is wrong. It outputs the full url, not a name. Should be RequestUri= like the C# sample.

context.log('Node.js HTTP trigger function processed a request. Name=' + req.originalUrl);

PR #33, opened 2016-03-14
Updating templates for recent binding changes

mathewc:
Absorbing changes from https://github.com/Azure/azure-webjobs-sdk-script/commit/81ee8b71b4f6c807ee2cbb0661685114fcd81a70

Also making a few additional cleanups/fixes. E.g., I'm adding some missing samples to the sln file. I added the sln as a way for us to have a project to open to facilitate working on samples. I realize some people use Sublime, but I don't, so keeping this in sync helps me alot. What are others using? Either we keep the sln in sync or remove it - I'm for keeping it :)

### DIFF
--- a/Templates/HttpTrigger/index.js
+++ b/Templates/HttpTrigger/index.js
@@ -1,5 +1,5 @@
 ﻿module.exports = function (context, req) {
-    context.log('Node.js HTTP trigger function processed a request. Name=' + req.originalUrl);
+    context.log('Node.js HTTP trigger function processed a request. RequestUri=' + req.originalUrl);
 
     if (typeof req.query.name == 'undefined') {
         context.res = {
--- a/Templates/QueueTrigger-Python/function.json
+++ b/Templates/QueueTrigger-Python/function.json
@@ -7,7 +7,7 @@
       },
       {
         "type": "table",
-        "name": "input",
+        "name": "tableInput",
         "direction": "in",
         "tableName": "samples",
         "partitionKey": "samples",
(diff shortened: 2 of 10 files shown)


Templates/HttpTrigger/index.js
```diff
@@ -1,5 +1,5 @@
 ﻿module.exports = function (context, req) {
-    context.log('Node.js HTTP trigger function processed a request. Name=' + req.originalUrl);
+    context.log('Node.js HTTP trigger function processed a request. RequestUri=' + req.originalUrl);
```
mathewc: Fix for https://github.com/Azure/azure-webjobs-sdk-templates/issues/31

davidebbo:
:shipit: But be aware that it goes live as soon as you merge. So if it's dependent on your script changes, let's not merge till those are live.

Templates/QueueTrigger-Python/function.json
```diff
@@ -7,7 +7,7 @@
       },
       {
         "type": "table",
```
mathewc: All of the json files have an extra level of unecessary indenting. Anyone know why, or whose editor is doing this?

fabiocav: I believe @soninaren is using VS Code with the default indentation configuration (4 spaces) and that has updated our JSON files.

mathewc:
I'll hold off committing this until we update, since as David says these updates rely on runtime updates.

Templates/QueueTrigger-Python/function.json
mathewc: So what is everyone else using, 4 spaces? Both my machines are using 2 (VS Enterprise) and I don't recall changing any defaults.

davidebbo: VS uses 4 spaces for me for JSON. Not sure what;s the default, but we should all use the same...

soninaren:
this has been merged in already.

Issue #29, opened 2016-03-12
Templates for Empty functions


paulbatum:
Lets start with C# and Node.js. The template would include an empty function signature but no bindings.

christopheranderson:
Looks like this is done now.

PR #35, opened 2016-03-14
Adding Empty C# and Node templates

soninaren:


soninaren:
For #29

soninaren:
@alrod to review if this will cause portal bugs

Templates/Empty-CSharp/function.json
```diff
@@ -0,0 +1,4 @@
+{
```
alrod: Why does not add "disabled": false ?

soninaren: being empty, wanted to leave the option to the user to disable or enable it

soninaren:
Talked to @alrod offline. He is okay merging this PR

Templates/Empty-CSharp/function.json
davidebbo: I think setting it explicitly to the default value makes it more usable, without taking anything away from it

Issue #39, opened 2016-03-15
Rename "broken" category to "experimental"


christopheranderson:
An upcoming update in the scripts repo will help with the "broken"-ness of the non-CS/JS languages. We should rename that category to experimental.

mathewc:
Yay :)

soninaren:
This is compelte

PR #42, opened 2016-03-15
Making category an array in metadata.json & adding experimental category to it

soninaren:


soninaren:
Addresses #39

davidebbo:
:shipit: but make sure Portal is able to consume it before getting it in (or at least that it doesn't completely blow up).

soninaren:
talked @alrod will merge once the portal is ready to consume the changes

PR #45, opened 2016-03-17
adding DocumentDB output binding

brettsam:


soninaren:
:+1: Looks good, Will talk to @alrod if there is something that needs to be done for portal

PR #71, opened 2016-03-25
Making template names consistent

mathewc:
Playing around in the UI, the inconsistencies are bothering me. Addressing some of them :)

### DIFF
--- a/Templates/QueueTrigger-Php/metadata.json
+++ b/Templates/QueueTrigger-Php/metadata.json
@@ -1,7 +1,7 @@
 {
     "defaultFunctionName": "QueueTriggerPhp",
     "description": "A PHP function that will be run whenever a message is added to a specified Azure Storage queue",
-    "name": "QueueTrigger Php Function",
+    "name": "QueueTrigger - Php",
     "language": "Php",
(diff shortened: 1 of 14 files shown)


Templates/QueueTrigger-Php/metadata.json
```diff
@@ -1,7 +1,7 @@
 {
     "defaultFunctionName": "QueueTriggerPhp",
     "description": "A PHP function that will be run whenever a message is added to a specified Azure Storage queue",
-    "name": "QueueTrigger Php Function",
+    "name": "QueueTrigger - Php",
```
davidebbo: Since you're in there, PHP is always upper case. Couple places in here.

christopheranderson:
Looks fine. I like the shorter names. I think this fits our beta milestone bar. :+1:

davidebbo:
:shipit: I've been wanting to do the same thing. Getting rid of 'function' makes the name shorter.

Templates/QueueTrigger-Php/metadata.json
mathewc: done

Issue #67, opened 2016-03-24
Remove runOnStartup from timer binding and template


paulbatum:
RunOnStartup has a problem where it will run every time that the function app gets loaded up due to other bindings (such as another function reading queue messages). This means that the function wont actually run according to the schedule. We need to remove it from the templates and the binding.json so that this option is disabled.

mathewc:
Yeah, TimerTrigger RunOnStartup is really intended for a long running host that doesn't restart often (as in classic WebJobs SDK scenarios). It's also useful for testing, which is why I think our templates inherited this from my initial samples.

PR #80, opened 2016-03-26
Removing runonstartup from timertrigger and binding.json

soninaren:


soninaren:
fixes #67

mathewc:
Looks good to me

PR #83, opened 2016-03-27
Simplifying C# HTTP trigger template

### LINKED ISSUE #63: HTTP C# template is too complicated
I do not think the current code looks approachable. This template probably needs to be rewritten from scratch.

fabiocav:
Resolves #63

### DIFF
--- a/Templates/HttpTrigger-CSharp/run.csx
+++ b/Templates/HttpTrigger-CSharp/run.csx
@@ -1,44 +1,22 @@
-#r "Newtonsoft.Json"
-
-using System;
 using System.Net;
 using System.Threading.Tasks;
-using Newtonsoft.Json;
 
-public static Task<HttpResponseMessage> Run(HttpRequestMessage req, TraceWriter log)
+public static async Task<HttpResponseMessage> Run(HttpRequestMessage req, TraceWriter log)
 {
-    // parse query parameter
-    var queryParam = req.GetQueryNameValuePairs()
-                 .ToDictionary(p => p.Key, p => p.Value, StringComparer.OrdinalIgnoreCase);
-
     log.Verbose($"C# HTTP trigger function processed a request. RequestUri={req.RequestUri}");
 
+    // parse query parameter
+    string name = req.GetQueryNameValuePairs()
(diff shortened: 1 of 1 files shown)


davidebbo:
Very nice! I'd take

```c#
    // Set name to query string, body data, falling back to some default
    name = name ?? data?.name ?? "Azure Functions";

    return req.CreateResponse(HttpStatusCode.OK, "Hello " + name);
```

fabiocav:
I tried to keep the same functionality we had before (and match Node), but I like that!

paulbatum:
This is definitely better, but I think I want to go a step further: is this functionality of asking for a query string parameter really what we want in the default template? I think I would prefer for both this template and the JS template to just return a 200 with a JSON response { message: "hello world" }

mathewc:
I think we go back and forth in these templates between trying to keep them clean/simple and trying to hint at some of the things people can do (e.g. showing APIs).

Whichever way we go, I think we should be consistent across all the templates. I'd be fine if we made them all super clean - just show how to access the input and write/return output, and leave it up to the user to refer to docs to learn about APIs/possibilities.

paulbatum:
OK lets go with Fabio's updated version for now, we can always revisit again later. :shipit:

Templates/HttpTrigger-CSharp/run.csx
```diff
@@ -1,44 +1,22 @@
...
-            Content = new StringContent("Hello " + name)
-        };
-    }
+    // Set name to query string or body data
+    name = name ?? data?.name;
 
-    return Task.FromResult(res);
-}
+    return name == null
+        ? req.CreateResponse(HttpStatusCode.BadRequest, "Please pass  a name on the query string or in the request body")
```
mathewc: Looks like there is an extra space between "pass" and "a"?

fabiocav:
Good eye! This was a simple string fix, so I just merged the PR.

Templates/HttpTrigger-CSharp/run.csx
fabiocav: Good eye! This was a simple string fix, so I just merged the PR.

mathewc: Thanks :)

PR #86, opened 2016-03-28
Add image resize sample for C#

lindydonna:
I've added image resizing based on the ImageResizer nuget. It is deployed as a private assembly since NuGet restore doesn't currently work for templates.

Putting in the experimental category for now, until we test in production.


Templates/BlobTrigger-ImageResize-CSharp/function.json
```diff
@@ -0,0 +1,22 @@
+{
```
soninaren: add "disabled": false,

Templates/BlobTrigger-ImageResize-CSharp/function.json
```diff
@@ -0,0 +1,22 @@
...
+    "bindings": [
+        {
+            "path": "samples/{name}",
```
mathewc: samples-images?

Templates/BlobTrigger-ImageResize-CSharp/function.json
```diff
@@ -0,0 +1,22 @@
...
+            "type": "blobTrigger",
+            "name": "input",
+            "direction": "in"
+        },
+        {
+            "path": "samples-sm/{name}",
```
mathewc: samples-images-sm?

Templates/BlobTrigger-ImageResize-CSharp/metadata.json
```diff
@@ -0,0 +1,11 @@
+{
+    "name": "Image resizing with a blob trigger - C#",
```
mathewc: This name doesn't match the naming scheme of all the other templates - this name looks more like a description. Wondering how well it will display in the UI.

Templates/BlobTrigger-ImageResize-CSharp/readme.md
```diff
@@ -0,0 +1,10 @@
+# Image resizing with a blob trigger - C<span>#</span>
+
+Using a blob trigger and the [ImageResizer](https://www.nuget.org/packages/ImageResizer/) nuget package, you can create a function that resizes new blobs when they are added to a container. This sample shows how to use a blob trigger to trigger the function and how to use output blobs with the same input blob name.
```
mathewc: "...that resizes new images..."

Templates/BlobTrigger-ImageResize-CSharp/run.csx
```diff
@@ -0,0 +1,49 @@
...
+    { ImageSize.Small, new ResizeSettings("maxwidth=640&maxheight=400") },
+    { ImageSize.Medium, new ResizeSettings("maxwidth=800&maxheight=600") }
+};
+
+#endregion
+
+public static void Run(
+    Stream input,                                     // input blob, large size
+    Stream blobOutputSmall, Stream blobOutputMedium,  // output blobs
+    TraceWriter log)
```
mathewc: Recommend removing TraceWriter if you're not using it

Templates/BlobTrigger-Simple-ImageResize-CSharp/function.json
```diff
@@ -0,0 +1,22 @@
+{
```
mathewc: Not sure we need both resize samples - can't we just have the one above?

lindydonna: Well, the problem is that the nuget one is not at all automated--users have to go into Kudu and manually touch project.json. Since image resizing seems like such a basic scenario, I would like customers to be able to do something easily.

Templates/BlobTrigger-ImageResize-CSharp/function.json
lindydonna: Good idea, I'll rename.

lindydonna:
@mathewc and @soninaren could you re-review and merge if it looks good?

Templates/BlobTrigger-ImageResize-CSharp/function.json
```diff
@@ -0,0 +1,23 @@
...
+        {
+            "path": "sample-images/{name}",
+            "type": "blobTrigger",
+            "name": "input",
+            "direction": "in"
+        },
+        {
+            "path": "sample-images-sm/{name}",
+            "type": "blob",
+            "name": "blobOutputSmall",
```
mathewc: nit: I'd rename these params simply imageSmall / imageMedium.

Templates/BlobTrigger-ImageResize-CSharp/function.json
```diff
@@ -0,0 +1,23 @@
+{
+    "disabled": false,
+    "bindings": [
+        {
+            "path": "sample-images/{name}",
+            "type": "blobTrigger",
+            "name": "input",
```
mathewc: might name this "image"

mathewc:
Looks good to me. I still think you should just add the one sample - I don't think we need the more complicated one.

lindydonna:
Since we're freezing changes right now, we'll merge next week. I'll fix the nits. :)

lindydonna:
@fabiocav had a great idea--we can distribute the ImageResizer binary directly because it is under MIT license. I'll revise the PR. And rebase so that the history looks nice. :)

soninaren:
@lindydonna we can merge this to dev, will automatically go in when dev is merged into master.

davidebbo:
I'm not a big fan of committing the imageresizer binary. Where does it come from? Can't we just get it from NuGet instead?

lindydonna:
No, because NuGet package restore does not currently work for templates. The only way to use it is for the customer to manually touch project.json in Kudu, which most customers won't be able to figure out. 

I'm not sure if there's something we can put into the build instead to download the NuGet, but this is only a temporary fix.

davidebbo:
But still, where does the exe come from? The problem with binaries in git is that they forever remain in the history even after you get rid of them.

Maybe this is an pushing factor to do things right with template NuGet restore. Is there a bug tracking that it doesn't work?

lindydonna:
My template only has a DLL, not exe. You might be thinking of Mathew's sample. 

Here's the bug: https://github.com/Azure/azure-webjobs-sdk-script/issues/196. Fabio said it's not that straightforward to fix, so I wanted to have a template before then--since image resizing is such a popular scenario.

davidebbo:
I guess we can bit the bullet on this one. As long as we don't make it a habit to commit exes/dlls in git! :)

lindydonna:
@davidebbo Agreed. Thanks for reviewing.

lindydonna:
@soninaren Should I open a new PR that merges into dev, or can you manually merge into dev?

soninaren:
@lindydonna I can merge manually into dev. Can you rebase on dev?

lindydonna:
@soninaren Sure, will update shortly.

lindydonna:
@soninaren Rebased onto dev.

soninaren:
@lindydonna dll file is breaking the portal because of the way it loads all the templates. i am doing additional testing to verify.

soninaren:
A bug in portal will not render this template correctly. Please don't merge until the portal bug is fixed. It will cause the portal to show errors or will not copy over the dll files in the template

paulbatum:
@lindydonna I'm afraid it looks like we gave you bad advice. The DLL approach won't work and we don't intend on fixing it - supporting binaries in templates introduces more complexity than its worth.

At this point I would just go with system.drawing with the caveats called out and then update the template once project.json support in templates is fixed.

lindydonna:
I'm closing this PR and opening a new one against the right branches.

PR #90, opened 2016-03-29
Make JS queue sample simpler. Fixes #89

### LINKED ISSUE #89: JS queue trigger template has confusing .id reference
The template has this code:

```
module.exports = function (context, myQueueItem) {
    context.log('Node.js queue trigger function processed work item', myQueueItem.id);
    context.done();
};
```
The confusing thing is that this .id reference is not referring to the actual azure queue id of the message.. its based on the sample data having an ID property.

To make this sample less confusing, lets change the sample data to something else.

paulbatum:


soninaren:
:+1: Looks good

Templates/QueueTrigger-NodeJS/index.js
```diff
@@ -1,4 +1,4 @@
 ﻿module.exports = function (context, myQueueItem) {
-    context.log('Node.js queue trigger function processed work item', myQueueItem.id);
```
mathewc: Should probably remove the "work item" nomenclature from these samples. I had that there because they used to be JSON work items with an ID.

I'd search for all "work item" (I used it in other samples) and just reword to "processed queue item"

Issue #104, opened 2016-03-29
Remove service bus bindings and templates


paulbatum:
These aren't in a good enough state yet. We can bring them back after beta, but we need to remove them now.

soninaren:
fixes #106

PR #105, opened 2016-03-29
removing service bus bindings and templates

soninaren:


soninaren:
fixes #104

soninaren:
@alrod can you review

alrod:
@soninaren, looks good

PR #111, opened 2016-03-30
[DO NOT MERGE] Updating copy for bindings

mattchenderson:
Updating copy. Not to be merged until signed off from UX team.

### DIFF
--- a/Bindings/bindings.json
+++ b/Bindings/bindings.json
@@ -2,9 +2,8 @@
     "$schema": "<TBD>",
     "contentVersion": "2016-03-04-alpha",
     "variables": {
-        "storageConnStringLabel": "Which storage account should be used?",
-        "appSettingsHelp": "You can use %&lt;App Setting Name&gt;% to reference an App Setting you've already configured",
-        "triggerRefHelp": "You can use brackets ({value}) and percent symbols (%value%) to format your path based on parameters from the trigger object."
+        "storageConnStringLabel": "Storage account connection",
+        "storageConnHelp": "This is the connection string for your storage account."
     },
     "bindings": [
         {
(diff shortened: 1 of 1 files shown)


Bindings/bindings.json
```diff
@@ -12,15 +13,22 @@
...
-                    "help": "A cron expression is a way of expressing a schedule. Please consult the documentation to learn more."
+                    "label": "Schedule",
+                    "help": "Enter a cron expression of the format '{second} {minute} {hour} {day} {month} {day of week}' to specify your schedule. Please consult the documentation to learn more."
+                },
+                {
+                    "name": "runOnStartup",
+                    "value": "boolean",
+                    "defaultValue": false,
+                    "label": "Should this function run on startup?",
+                    "help": "When true, your function will run on startup. This may be useful for debugging."
```
davidebbo: Sound like this section shouldn't be there?

Bindings/bindings.json
```diff
@@ -2,7 +2,8 @@
     "$schema": "<TBD>",
     "contentVersion": "2016-03-04-alpha",
     "variables": {
-        "storageConnStringLabel": "Which storage account should be used?",
+        "storageConnStringLabel": "Storage account connection",
+        "storageConnHelp": "This is the connection string for your storage account.",
         "appSettingsHelp": "You can use &lt;App Setting Name&gt; to reference an App Setting you've already configured",
```
davidebbo: Looks like `appSettingsHelp` is no longer used at all. I would just stay with `appSettingsHelp` and keep using it everywhere instead of introducing new variable. This way the change can really be 'just strings'.

davidebbo:
I created a new cleaner PR for it: #112

PR #118, opened 2016-03-31
Removing Queue trigger F# template

soninaren:


davidebbo:
We really should have a 'disable' flag that we can set to make the UI ignore it without yanking everything.

But for now this will do!

soninaren:
@alrod can you do a quick signoff on this

mathewc:
Yes I was thinking the same thing. Logged: https://github.com/Azure/azure-webjobs-sdk-templates/issues/119

alrod:
@soninaren, looks good

soninaren:
:+1:

PR #122, opened 2016-03-31
Fixing and reenabling ServiceBus templates

mathewc:


Templates/ServiceBusQueueTrigger-NodeJS/index.js
```diff
@@ -0,0 +1,4 @@
+﻿module.exports = function(context, myQueueItem) {
+    context.log('Node.js ServiceBus queue trigger function processed message', myQueueItem);
+    context.done(null, OutPutQueueItem);
```
soninaren: This should be `context.done()` since there is not output binding

mathewc: Good catch - I'm testing it in prod and already caught that. I'll make sure it runs e2e.

mathewc: Fixed.

soninaren:
:+1:  Looks good

PR #123, opened 2016-04-01
Adding GithubWebHook-CSharp template

soninaren:


Templates/GitHubWebHook-CSharp/run.csx
```diff
@@ -0,0 +1,15 @@
+using System.Net;
+using System.Threading.Tasks;
+
+public static async Task<HttpResponseMessage> Run(HttpRequestMessage req, TraceWriter log)
+{
+    log.Verbose($"C# HTTP trigger function processed a request. RequestUri={req.RequestUri}");
+
+    // Get request body
+    dynamic data = await req.Content.ReadAsAsync<object>();
```
davidebbo: We have a strongly typed object from ASP.NET WebHooks library. I think it would be better to use that instead of dynamic.

soninaren: Do you mean something like this?  
```
JObject data = await req.Content.ReadAsAsync<JObject>();
string gitHubComment = data?["comment"]?["body"].ToString();
```
From what i see in the Script SDK the httprequest is passed on to the function from WebHookHandlerContext

davidebbo: I was expecting the WebHooks library to have a class for the GitHub payload, but it doesn't look like it (@HenrikFrystykNielsen?).

So go with what you had initially with dynamic. We can always evolve later.

davidebbo:
:shipit:

soninaren:
:+1:

PR #125, opened 2016-04-01
Revert "Removing Queue trigger F# template"

paulbatum:
This reverts commit 3f3a3860fbf60c0d418b2732ab106514a35d437a.

### DIFF
--- a/Templates/QueueTrigger-FSharp/metadata.json
+++ b/Templates/QueueTrigger-FSharp/metadata.json
@@ -1,0 +1,12 @@
+{
+    "defaultFunctionName": "QueueTriggerFSharp",
+    "description": "An F# function that will be run whenever a message is added to a specified Azure Storage queue",
+    "name": "QueueTrigger - F#",
+    "language": "F#",
+    "trigger": "QueueTrigger",
+    "category": [ "Experimental" ],
+    "userPrompt": [
+        "connection",
+        "queueName"
+    ]
+}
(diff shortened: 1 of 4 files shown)


Templates/QueueTrigger-FSharp/metadata.json
```diff
@@ -0,0 +1,12 @@
+{
+    "defaultFunctionName": "QueueTriggerFSharp",
+    "description": "An F# function that will be run whenever a message is added to a specified Azure Storage queue",
+    "name": "QueueTrigger - F#",
+    "language": "F#",
+    "trigger": "QueueTrigger",
```
mathewc: "trigger" property is obsolete

mathewc:
Thanks :) :shipit:

Templates/QueueTrigger-FSharp/metadata.json
soninaren: The end 2 end script uses it. Lets keep it for now until i update the script.

paulbatum: I believe the trigger metadata is being used by the CI automation, Naren can confirm. I double checked across the templates and there is some inconsistency between which ones have it and which ones dont. I will leave it here right now, we can do another pass to get this in order.

soninaren: Sorry. E2E uses it from function.json. We can remove from here

paulbatum: Ahh that makes sense.

soninaren:
:+1: Looks good

Issue #128, opened 2016-04-04
Make input and output default parameters names different


christopheranderson:
DocDB in and out binding default parameter names are the same, which often results in duplicate names. Let's use unique names for these.

From: https://github.com/Azure/azure-webjobs-sdk-script/issues/190

PR #131, opened 2016-04-06
renaming the default parameter for bindings

soninaren:


soninaren:
fixes #128

soninaren:
need to update more bindings adding additional commits

soninaren:
updated

soninaren:
@mattchenderson  or @christopheranderson can you review?

Bindings/bindings.json
```diff
@@ -268,7 +268,7 @@
             "type": "serviceBus",
             "displayName": "Azure Service Bus",
             "direction": "out",
-            "defaultParameterName": "myQueueItem",
+            "defaultParameterName": "sbOutputQueueItem",
```
christopheranderson: outputSbQueueItem would be more consistent

Bindings/bindings.json
```diff
@@ -243,7 +243,7 @@
             "type": "serviceBusTrigger",
             "displayName": "Azure Service Bus",
             "direction": "trigger",
-            "defaultParameterName": "myQueueItem",
+            "defaultParameterName": "sbQueueItem",
```
christopheranderson: inputSbQueueItem would be more consistent

christopheranderson:
Approved

Bindings/bindings.json
soninaren: following the format for trigger parameter with prefix "my"

Issue #98, opened 2016-03-29
ServiceBusQueue Node template should have a trigger only


fabiocav:
#### Expected behavior
When using the ServiceBusQueueTrigger template, I have a function with a trigger only

#### Actual behavior
The current template has an trigger and an output binding

fabiocav:
Resolved by #127

PR #132, opened 2016-04-06
Adding row key setting for table storage input binding. Updating table and blob bindings.

soninaren:
…


soninaren:
fixes #98

christopheranderson:
Tested rand-guid. 

Approved

Issue #136, opened 2016-04-07
Regex for blob trigger name validation should allow only container names


soninaren:
Only contianer names are allowed for the path setting of blob trigger. Since the trigger could go off whenever there is a new blob added to the container

soninaren:
This is merged

PR #137, opened 2016-04-08
Fixed the regex for blob binding

soninaren:
fixing the trigger regex to allow container + blob name as well as only blob name
Fixed the input and output regex to allow only one consecutive - in the container name


soninaren:
fixes #136

christopheranderson:
@alrod - could you review that this fixes the issue? Once you sign off, this is approved.

alrod:
looks good

Issue #50, opened 2016-03-20
Blob trigger node.js - change path name from sample-workitems to myblob


nirmash:
not a huge priority but couple of issues with current name:
1. sample-workitems sounds more like a queue not a file folder 
 2. having a dash is illegal for table storage (even though it is ok for blob storage might as well make it consistent) 
3. new name will be similar to the name of the object in code

mathewc:
Changing the name is fine, but not "myblob" since the path is a **container** for multiple blobs. Perhaps "my-container" or "sample-items" or something like that. Also, the casing/naming of queues/containers shouldn't be camel cased like the parameter name (which is that way because it's a code variable).

All these "sample" names came from my initial sample scripts in the repo, where I was scenario driven, and the scenario for queues/blobs was that JSON work items were added and the scripts processed them.

christopheranderson:
This is being done in the PR above.

soninaren:
this is merged

PR #138, opened 2016-04-08
updated default settting values for all bindings

soninaren:


soninaren:
fixes #50

christopheranderson:
This is much cleaner. Thanks for these changes. As long as no one else has feedback, this is approved. :shipit:

PR #145, opened 2016-04-13
Add authLevel setting for Http Trigger

christopheranderson:
This enables the authLevel config for Http Trigger.

I've tested these settings on a production Function App.

The original enum can be found here: https://github.com/Azure/azure-webjobs-sdk-script/blob/master/src/WebJobs.Script/AuthorizationLevel.cs

Note that we're ignoring the `User` setting for now since it doesn't work as intended.

### DIFF
--- a/Bindings/bindings.json
+++ b/Bindings/bindings.json
@@ -274,6 +274,7 @@
                 {
                     "name": "authLevel",
                     "value": "enum",
+                    "defaultValue": "function",
                     "enum": [
                         {
                             "value": "function",
--- a/Bindings/bindings.json
+++ b/Bindings/bindings.json
@@ -270,6 +270,26 @@
                     ],
                     "label": "WebHook type",
                     "help": "WebHooks allow services to call one another easily. If the type you need isn't listed here, choose 'Not a WebHook'."
+                },
+                {
+                    "name": "authLevel",
+                    "value": "enum",
+                    "enum": [
+                        {
+                            "value": "anonymous",
+                            "display":"Anonymous"
+                        },
+                        {
+                            "value": "function",
+                            "display":"Function"
+                        },
+                        {
+                            "value": "admin",
+                            "display": "Admin"
+                        }
+                    ],
+                    "label": "Authorization level",
+                    "help": "Authorization level controls whether you're using an API key and which key to use. The function & master keys are found in your host.json. For user-based authentication, go to Function App Settings."
                 }
             ]
         },
--- a/Bindings/bindings.json
+++ b/Bindings/bindings.json
@@ -276,12 +276,12 @@
                     "value": "enum",
                     "enum": [
                         {
-                            "value": "anonymous",
-                            "display":"Anonymous"
-                        },
-                        {
                             "value": "function",
                             "display":"Function"
+                        },
+                        {
+                            "value": "anonymous",
+                            "display":"Anonymous"
                         },
                         {
                             "value": "admin",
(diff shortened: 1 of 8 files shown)


Bindings/bindings.json
```diff
@@ -270,6 +270,26 @@
...
+                            "value": "function",
+                            "display":"Function"
+                        },
+                        {
+                            "value": "admin",
+                            "display": "Admin"
+                        }
+                    ],
+                    "label": "Authorization level",
+                    "help": "Authorization level controls whether you're using an API key and which key to use. The function & master keys are found in your host.json. For user-based authentication, go to Function App Settings."
```
mathewc: How is the default controlled? It should remain "function" (that is the default when it isn't specified).

mathewc: Might need to clarify that the host.json isn't the regular host.json, it's the host.hson secrets file. Also, might need to mention that the master key maps to admin level (since two different terms are being used).

soninaren:
Tested in prod portal. Works fine :+1:

christopheranderson:
We've found a portal bug with this, so it's not ready to merge. I'll follow up with a portal bug number...

Bindings/bindings.json
christopheranderson: Good point. I've updated the new one. Default is the "first" value, but we've found a cool bug here, so it might change.

christopheranderson: Nope. We found the issue.

christopheranderson:
I've added fixes for Mathew's comments and the default value issue. @soninaren please validate.

soninaren:
It works when we add a new trigger. Does work fine when we use an existing template. Will a have to update all the http templates with default value.

Bindings/bindings.json
```diff
@@ -270,6 +270,27 @@
...
+                            "value": "anonymous",
+                            "display":"Anonymous"
+                        },
+                        {
+                            "value": "admin",
+                            "display": "Admin"
+                        }
+                    ],
+                    "label": "Authorization level",
+                    "help": "Authorization level controls whether you're using an API key and which key to use; Function uses your function key; Admin uses your master key. The function & master keys are found in your host secrets file. For user-based authentication, go to Function App Settings."
```
mathewc: nit: your use of ';' as separators in this sentence seems off. Recommend rewording as a proper sentence :)

soninaren:
Unfortunately with default value set as function. And function being the first value in the dropdown the save button doesn't appear for http trigger. Portal will have to fix this bug. Filling an issue against portal for this.

Templates/GenericWebHook-CSharp/function.json
```diff
@@ -5,6 +5,7 @@
             "type": "httpTrigger",
             "direction": "in",
             "webHookType": "genericJson",
+            "authLevel": "function",
```
mathewc: authLevel only applies to non WebHook functions. WebHooks by definition use their own auth. So I don't think you should show this authLevel for WebHook functions, and you shouldn't have this default value in there. See controller auth code [here](https://github.com/Azure/azure-webjobs-sdk-script/blob/master/src/WebJobs.Script.WebHost/Controllers/FunctionsController.cs#L60).

soninaren: Right now webhook doesn't have a separate trigger type. So portal will have to have a mechanism to only show auth level setting when webhook type is selected.

christopheranderson: This is essentially the same problem as Service Bus Topic name vs Queue name. I believe Alexey is adding support for that now. I'll pull that commit out and we can add the "exclusive" rule to the binding.json. We'll need to hold this until we fix the enum portal issues anyway.

soninaren:
opened another PR using exclusivity rule. https://github.com/Azure/azure-webjobs-sdk-templates/pull/156

PR #147, opened 2016-04-14
Removing the rowkey from table input binding.

soninaren:


fabiocav:
This looks good. 👍

mathewc:
There might be valid scenarios where you want to input bind to a single row, right?

The Table binding needs work and we have issue(s) tracking that. However, I believe now that the binding supports both query based bindings (returning an array) as well as a singleton?

soninaren:
adding rowkey is not working at the moment. Have removed it temporarily. will add it back once the issue has been resolved.

https://github.com/Azure/azure-webjobs-sdk-script/issues/261

mathewc:
Got it. I'll handle that issue with the batch of other Table binding work I'll be getting to soon.

Issue #117, opened 2016-03-31
Review/Simplify all the Experimental templates


mathewc:
We should just simplify them so they just respond to the input trigger, not include all kinds of binding chains. E.g. some get a queue event, read from a table, then write to blob. These operations are carry overs from my initial sample templates.

PR #149, opened 2016-04-15
Cleaning up templates in experimental category

soninaren:


soninaren:
fixed #117

Templates/QueueTrigger-Python/function.json
```diff
@@ -15,13 +15,7 @@
...
             "partitionKey": "samples",
             "take": 5,
-            "filter": "status eq '0'"
-        },
-        {
-            "type": "blob",
-            "name": "output",
-            "direction": "out",
-            "path": "samples-output/{id}"
+            "filter": "true"
```
mathewc: Recommend removing this noop filter

Templates/QueueTrigger-Powershell/queueTrigger.ps1
```diff
@@ -1,8 +1,6 @@
 $in = Get-Content $Env:input
+$output = $Env:output
 
 [Console]::WriteLine("Powershell script processed queue message '$in'")
-
-$output = $Env:output
-$json = $in | ConvertFrom-Json
-$entity = [string]::Format('{{ "timestamp": "{0}", "status": 0, "title": "Powershell Table Entity for message {1}" }}', $(get-date -f MM-dd-yyyy_HH_mm_ss), $json.id)
+$entity = [string]::Format('{{ "timestamp": "{0}", "title": "queue message:{1}" }}', $(get-date).ToString(), $in)
```
mathewc: Missing space in "queue message:{1}"

mathewc:
These changes don't really seem to address #117. The idea in that issue was to make them simple, so they're not a mashup of various bindings. E.g., for these, maybe we just have them get the queue message and that is it.

What do others think? I think we should rely on doc/samples to show the possibilities for various languages, not the template bindings.

soninaren:
The reason i did not remove table binding from these templates is because these are the only 2 templates with table bindings. I feel we should have at least one sample of table input and output binding until we add more.

mathewc:
Ok.

soninaren:
Addressed the PR comments

Issue #134, opened 2016-04-07
Split Service Bus binding into two


fabiocav:
Currently, the service bus binding only exposes a parameter for the queue, and since the portal doesn't have a mechanism to pick between two required parameters, we should split the binding template into two, one for queues (Service Bus Queue) and one for topics (Service Bus Topic)

christopheranderson:
So we have two options for changing this:

1. We add a new "type" to the Script Host for "serviceBusTopic". Mathew doesn't like this and he has a point that we need to support a similar thing for API Hubs eventually. This would require some pretty large script host changes.

2. We allow for the portal to work off of more than just "type" and "direction" for changing the UX. We could add a "rule" property which lets us set things like "exclusivity" between settings. (and potentially more rules, which I can see coming our way with API Hubs....). This would require some pretty large portal changes.

Thoughts? 

Example bindings.json snippet:

```json
{
    "type": "serviceBusTrigger",
    "displayName": "Azure Service Bus",
    "direction": "trigger",
    "defaultParameterName": "myQueueItem",
    "parameterNamePrompt": "Message parameter name",
    "settings": [
        {
            "name": "queueName",
            "value": "string",
            "defaultValue": "samples-input",
            "required": true,
            "label": "Queue name",
            "help": "This is the name of the queue that messages will be read from."
        },
        {
            "name": "topicName",
            "value": "string",
            "defaultValue": "samples-input",
            "required": true,
            "label": "Topic name",
            "help": "This is the name of the queue that messages will be read from."
        },
        {
            "name": "connection",
            "value": "string",
            "resource": "ServiceBus",
            "required": true,
            "label": "Service Bus connection",
            "help": "This is the connection string for your Service Bus account. Make sure that it has Listen rights."
        }
    ],
    "rules": [
        {
            "type": "exclusivity",
            "values": [
                {
                    "value": "queueName",
                    "displayName": "Service Bus Queue"
                },
                {
                    "value": "topicName",
                    "displayName": "Service Bus Topic"
                }
            ],
            "label": "Service Bus message type",
            "help": "This is the type of Servce Bus message. This lets you choose between using Queues or Topics."
        }
    ]
},
```

alrod:
I have some changes in proposed format:

{
            "type": "serviceBusTrigger",
            "displayName": "Azure Service Bus",
            "direction": "trigger",
            "defaultParameterName": "myQueueItem",
			"settings": [
				{
					"name": "queueName",
[...]

PR #153, opened 2016-04-18
Fixing service bus binding.

soninaren:


soninaren:
fixes #134 
@alrod can you take a look?

alrod:
looks good

PR #156, opened 2016-04-19
Adding AuthLevel to http trigger using exclusivity rule.

soninaren:


soninaren:
@alrod , @christopheranderson and @mattchenderson can you guys review?

Bindings/bindings.json
```diff
@@ -269,7 +265,57 @@
...
+                            "hiddenSettings": [
+                                "authLevel"
+                            ],
+                            "shownSettings": [
+                                "webHookType"
+                            ]
+                        }
+                    ],
+                    "label": "Mode",
+                    "help": "Mode of trigger. Standard mode means its a normal http trigger, WebHook mode means it is a http trigger of webhook type."
```
mattchenderson: This is the mode of the trigger. "Standard" means that the request will be standard HTTP with no additional semantics. "Webhook" means that the request will be processed according to a specified webhook type.

PR #162, opened 2016-04-25
Adding manual trigger for C#

soninaren:


Templates/ManualTrigger-CSharp/metadata.json
```diff
@@ -0,0 +1,8 @@
+{
+    "defaultFunctionName": "ManualTriggerCSharp",
+    "description": "A C# function without triggers, inputs, or outputs",
```
mathewc: Let's update the description of this and the Node one to be something like:

"A C# function that is triggered manually via the portal"

Since it's confusing to say "without triggers" when it is called "Manual Trigger"

Templates/ManualTrigger-CSharp/metadata.json
```diff
@@ -0,0 +1,8 @@
...
+    "name": "ManualTrigger - C#",
+    "trigger": "MaunualTrigger",
+    "language": "C#",
+    "category": [ "Experimental" ]
```
mathewc: Both of these templates should be moved to the "Core" category, peer to the Empty template which is also Core

Templates/ManualTrigger-CSharp/run.csx
```diff
@@ -0,0 +1,5 @@
+using System;
```
mathewc: Looks like this file is missing a space between the usings and the code?

Templates/ManualTrigger-CSharp/sample.dat
```diff
@@ -0,0 +1 @@
+Foobar
```
mathewc: Let's not use "Foobar". How about something like "Test Value".

Also, I don't see a sample.dat for the Node ManualTrigger template?

Templates/ManualTrigger-CSharp/metadata.json
soninaren: Had earlier moved them to experimental because of this issue #87. Moving both of them to core.

soninaren:
@mathewc updated the PR code as per comments

Templates/ManualTrigger-CSharp/metadata.json
```diff
@@ -0,0 +1,8 @@
+{
+    "defaultFunctionName": "ManualTriggerCSharp",
+    "description": "A C# function that is triggered manually via the portal \"Run\" button",
+    "name": "ManualTrigger - C#",
+    "trigger": "MaunualTrigger",
```
mathewc: This trigger value is mispelled - isn't this failing tests?

soninaren: Don't know where it got misspelled, I tested manually before sending the PR. There is no automation setup for manual triggers at the moment, also the tests run after the PR has merged in.

mathewc: It's spelled "MaunualTrigger" above - just fix that before checking in.

Templates/ManualTrigger-NodeJS/metadata.json
```diff
@@ -1,8 +1,8 @@
 {
     "defaultFunctionName": "ManualTriggerNodeJS",
-    "description": "A Node.JS function without triggers, inputs, or outputs",
+    "description": "A Node.js function that is triggered manually via the portal \"Run\" button",
     "name": "ManualTrigger - Node",
     "trigger": "MaunualTrigger",
```
mathewc: Mispelled here as well

Templates/ManualTrigger-CSharp/metadata.json
soninaren: i meant don't know when it got missplled :)

mathewc: Uhhh, you hit the wrong keys on the keyboard maybe? :)

mathewc:
Thanks for taking care of this so quickly. After fixing trigger spelling, get it in :) :ship:

soninaren:
:+1:

PR #165, opened 2016-04-26
Adding UI for subscription Name for service bus topic binding.

soninaren:


Bindings/bindings.json
```diff
@@ -455,6 +455,20 @@
                     ]
                 },
                 {
+                    "name": "subscriptionName",
+                    "value": "string",
+                    "defaultValue": "mysubscription",
+                    "required": true,
+                    "label": "Subscription name",
+                    "help": "This is the name of the subscription from which the message will be read.",
```
mathewc: Recommend using the same or similar wording that the ServiceBusTriggerAttribute itself uses for these parameters

Bindings/bindings.json
```diff
@@ -455,6 +455,20 @@
                     ]
                 },
                 {
+                    "name": "subscriptionName",
+                    "value": "string",
+                    "defaultValue": "mysubscription",
```
mathewc: Use correct camel casing for variable names

mathewc: What purpose does this default value serve? It will never be valid, right?

Bindings/bindings.json
```diff
@@ -455,6 +455,20 @@
...
+                    "required": true,
+                    "label": "Subscription name",
+                    "help": "This is the name of the subscription from which the message will be read.",
+                    "validators": [
+                        {
+                            "expression": "^[0-9a-z][a-z0-9_.-]{1,48}[0-9a-z]$",
```
mathewc: Where did you get this regex from - does it map to the actual naming rules for ServiceBus subscription names?

soninaren: Got it from the old portal, It should allow uppercase characters though will update it.

Bindings/bindings.json
soninaren: what wording are you referring to?

mathewc: https://github.com/Azure/azure-webjobs-sdk/blob/master/src/Microsoft.Azure.WebJobs.ServiceBus/ServiceBusTriggerAttribute.cs#L57

soninaren:
updated the PR.

mathewc:
:ship:

Issue #157, opened 2016-04-20
ServiceBus Topic Template


mathewc:
I recently responded to a SO post where a user was trying to trigger a function on a ServiceBus topic (https://stackoverflow.com/questions/36707496/how-to-use-a-servicebus-trigger-with-a-topic-subscription-in-an-azure-function). The runtime supports this, but he had to go to the advanced editor to enable make it work.

Recommend creating a separate template for this. Also there might be portal work that needs to be done to expose UI for setting the topic/subscription.

soninaren:
Portal work for Service Bus Binding UI is done. It is available in the next portal at the moment. I will create a service bus topic template for the templates repo.

PR #168, opened 2016-04-27
Adding a new service bus topic template and update servicebus queue template

soninaren:


Templates/ServiceBusTopicTrigger-CSharp/function.json
```diff
@@ -0,0 +1,14 @@
+﻿{
+    "disabled": false,
+    "bindings": [
+        {
+            "name": "mySbMsg",
+            "accessRights" : "Manage",
```
mathewc: Can we not specify this if it has the default value?

Templates/ServiceBusTopicTrigger-NodeJS/function.json
```diff
@@ -0,0 +1,14 @@
+﻿{
+    "disabled": false,
+    "bindings": [
+        {
+            "name": "mySbMsg",
+            "accessRights" : "Manage",
+            "type": "serviceBusTrigger",
+            "direction": "in",
+            "topicName": "myinputtopic",
```
mathewc: Can you use the same topic name as the C# template? Why the arbitrary difference?

Templates/ServiceBusTopicTrigger-CSharp/function.json
```diff
@@ -0,0 +1,14 @@
...
+            "type": "serviceBusTrigger",
+            "direction": "in",
+            "topicName": "mytopic",
```
mathewc: Make your use of camel casing consistent throughout. In this case your topicName is not camel cased, but subscription is. Recommend checking the rest.

Templates/ServiceBusTopicTrigger-CSharp/function.json
soninaren: The drop down for the setting will be presented to the user at the time of creating the function. This value is going to be replaced by what user selects in the drop down anyway. I can remove this.

Templates/ServiceBusTopicTrigger-CSharp/metadata.json
```diff
@@ -0,0 +1,15 @@
...
+    "description": "A C# function that will be run whenever a message is added to a specified Service Bus topic",
+    "language": "C#",
+    "category": [
+        "Core",
+        "Data Processing"
+    ],
+    "userPrompt": [
+        "connection",
+        "accessRights",
+        "topicName"
```
mathewc: Why isn't subscriptionName also a user prompt?

Templates/ServiceBusTopicTrigger-CSharp/metadata.json
```diff
@@ -0,0 +1,15 @@
+{
+    "name": "ServiceBusTopicTrigger - C#",
+    "defaultFunctionName": "ServiceBusTopicTriggerCSharp",
+    "description": "A C# function that will be run whenever a message is added to a specified Service Bus topic",
```
mathewc: here and below, "the specified Service Bus topic"

Templates/ServiceBusTopicTrigger-CSharp/function.json
soninaren: Uppercase characters are not allowed in topic names. Hence the topic name in lowercase.

Templates/ServiceBusTopicTrigger-CSharp/metadata.json
soninaren: It should be, will fix this.

Templates/ServiceBusTopicTrigger-CSharp/function.json
mathewc: Then why not make the subscription lower as well? Aren't all the other default values for the various templates lower?

soninaren: I have used camel case for default values where uppercase characters are allowed for better readability.

Templates/ServiceBusTopicTrigger-CSharp/metadata.json
mathewc: You're trying these templates out end to end, right? The above shouldn't have worked.

soninaren: Yes i tested it out end to end. It worked because all mandatory settings are populated for the user from the default value in binding.json.

Templates/ServiceBusTopicTrigger-NodeJS/function.json
soninaren: Followed the pattern for storage queue templates. Which doesn't apply to topic because a topic can have multiple subscribers. For queue i did not want 2 competing functions if the user simply clicked create without modifying the queue name.

soninaren:
Updated as per comments.

soninaren:
Fixes #157

Templates/ServiceBusQueueTrigger-CSharp/function.json
```diff
@@ -2,7 +2,7 @@
     "disabled": false,
     "bindings": [
         {
-            "name": "myQueueItem",
+            "name": "mySbMsg",
```
mathewc: Why are you renaming these? The old names seem better to me.

soninaren: Matching with binding.json default values.

mathewc:
Also, add your new templates to the VS solution

Templates/ServiceBusQueueTrigger-CSharp/run.csx
```diff
@@ -1,7 +1,7 @@
 using System;
 using System.Threading.Tasks;
 
-public static void Run(string myQueueItem, TraceWriter log)
+public static void Run(string mySbMsg, TraceWriter log)
 {
-    log.Verbose($"C# ServiceBus queue trigger function processed message: {myQueueItem}");
+    log.Verbose($"C# ServiceBus queue trigger function processed message: {mySbMsg}");
```
mathewc: change your log statements to Info level rather than Verbose (see my open PR on this repo)

soninaren:
Updated

mathewc:
:ship:

PR #173, opened 2016-04-28
Renamed easy table to mobile table

soninaren:


Bindings/bindings.json
```diff
@@ -1065,7 +1065,7 @@
         },
         {
             "type": "easyTable",
```
mathewc: why is the type still easyTable? We've already released the breaking change to rename that to mobileTable in the runtime.

soninaren:
updated.

brettsam:
This looks good to me. Let's wait until 0.2 is the default version, then get this in.

soninaren:
This is merging into dev, Will merge dev to master once we have 0.2 as default.

PR #174, opened 2016-04-28
hide the apiHubFile binding from prod and staging

soninaren:


Bindings/bindings.json
```diff
@@ -383,6 +383,9 @@
                         "capability": "blob"
                     }
                 }
+            ],
+            "filters": [
+                "showHiddenBinding"
```
Tuesdaysgreen: In your email you said that you're hiding it from prod and staging, but this filter says "showHiddenBinding".  Doesn't that contradict what you said?  Regardless, we actually want to expose the API Hub File bindings so that we can start getting feedback on them.

soninaren: showHiddenBinding is the query string key you need to add show ApiHub Binding. It is hidden by default when the filter setting is present. My understanding was that we wanted hide API Hub File bindings.

galiniliev: Actually we want to release those. Please do not hide those

soninaren: :+1:

Issue #176, opened 2016-05-02
New template: Timer Trigger PowerShell


christopheranderson:
We need a new template for Timer Trigger for the PowerShell language.

PR #178, opened 2016-05-03
Adding powershell TimerTrigger template

soninaren:


soninaren:
#176

alrod:
Where is ps1 file?

soninaren:
https://github.com/soninaren/azure-webjobs-sdk-templates/blob/powerShelltimer/Templates/TimerTrigger-Powershell/run.ps1

Templates.sln
```diff
@@ -230,6 +230,16 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "ServiceBusTopicTrigger-Node
 		Templates\ServiceBusTopicTrigger-NodeJS\metadata.json = Templates\ServiceBusTopicTrigger-NodeJS\metadata.json
 	EndProjectSection
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "TimerTrigger-Powershell", "TimerTrigger-Powershell", "{6B9E2925-C92E-43C2-96FE-8A6E0C5C7E52}"
+	ProjectSection(SolutionItems) = preProject
```
mathewc: Thanks for keeping the sln up to date - I'm having to go in and add things all the time that people missed :)

Templates/TimerTrigger-Powershell/metadata.json
```diff
@@ -0,0 +1,13 @@
+{
+    "defaultFunctionName": "TimerTriggerPowershell",
+    "description": "A Powershell function that will be run on a specified schedule",
+    "name": "TimerTrigger - Powershell",
+    "language": "Powershell",
+    "category": [
+        "Experimental",
+        "Data Processing"
```
mathewc: It's not really data processing, right? I see the other Timer samples are in the same category, but they're also in Core. Let's at least make sure it's consistent across all the Timer templates.

soninaren: There is was not a very string argument against putting time trigger templates is data processing. The closes probable use of time triggers would be to fetch or process data periodically. In time we would have a schedule category in templates. All C# and Node templates are there in core unless the binding itself needs work.

mathewc: Doesn't this template belong in Core as well then, if the other Timer templates are?

soninaren: You are right. I need to either remove this from data processing and leave it just in experimental or have the other experimental templates show up in respective categories as well.
@christopheranderson , @mattchenderson  thoughts?

mathewc: I'd say just leave this in Experimental like all the others. It's not our intent at this time to promote the Powershell templates. If we did, we'd want to do it for the other Powershell templates as well.

soninaren: Makes sense. Removing this from Data Processing.

soninaren: Done

mathewc:
:ship

Issue #177, opened 2016-05-02
New template: HTTP Trigger PowerShell


christopheranderson:
We need a new template for HTTP Trigger in PowerShell

mattchenderson:
@alrod - did you say on email that you would do this one?

PR #181, opened 2016-05-03
Adding HttpTrigger Powershell template

soninaren:


soninaren:
#177

Templates/HttpTrigger-Powershell/run.ps1
```diff
@@ -0,0 +1,8 @@
+$name = Get-Content $Env:req
```
mathewc: This is so you can support POST as well?

Templates/HttpTrigger-Powershell/function.json
```diff
@@ -0,0 +1,16 @@
+{
+    "disabled": false,
+    "bindings": [
+        {
+            "authLevel": "function",
```
mathewc: This is the default auth level - doesn't need to be specified. Our other http samples don't specify it.

Templates/HttpTrigger-Powershell/run.ps1
soninaren: Yes, For consistency across HttpTrigger templates

Templates/HttpTrigger-Powershell/function.json
soninaren: will remove this.

Templates/HttpTrigger-Powershell/run.ps1
mathewc: Well the templates aren't really consistent - e.g. the BAT template only processes the name from the query string, otherwise returns "Enter a name"

soninaren: I will update batch template as well to make it consistent. However i am not sure how to return 404 for a non Node/C# template. So i am returning hello in case nothing is passed via post body or query string.

mathewc: How about coding this like the BAT template:

if ($Env:req_query_name) 
{
	$res = "Hello $Env:req_query_name"
}
else
{
	$res = "Please pass a name on the query string"
}

Out-File -NoNewline $Env:res -InputObject $res

soninaren: We wanted the Run button from portal to give a result with the request body. Originally the Node and C# were written this way as well. But we changed them.

mathewc: So why don't we make things consistent one way or another?

PR #182, opened 2016-05-03
Changed webhook to lowercase

soninaren:


soninaren:
Fixes #182 
@alrod can you take a look?

Issue #187, opened 2016-05-10
Todo List template using table storage for Node.js


soninaren:

soninaren:
Removing this one as per discussion on #192

PR #192, opened 2016-05-11
Adding Todo list template for Node.js

soninaren:


soninaren:
fixes #187

Templates/TodoList-NodeJS/metadata.json
```diff
@@ -0,0 +1,16 @@
+{
+    "defaultFunctionName": "TodoListNodeJS",
+    "description": "A Node.js function that maps http methods to CRUD operation on a Storage Table",
```
mathewc: CRUD operations

Templates/TodoList-NodeJS/function.json
```diff
@@ -0,0 +1,33 @@
+﻿{
+  "bindings": [
+    {
+      "authLevel": "function",
+      "type": "httpTrigger",
```
mathewc: Since you only support GET and POST, recommend adding methods: [ "get", "post" ]. The runtime will then restrict.

Templates/TodoList-NodeJS/index.js
```diff
@@ -0,0 +1,45 @@
+﻿module.exports = function (context, req, inputList) {
+    var method = req.method.toLowerCase();
+
+    if (methods[method]) {
+        methods[method](context, req, inputList);
+    } else {
+        context.res = { status: 404, body: "No route found for VERB: " + req.method };
```
mathewc: Once you put your methods restriction in the binding, you won't need this validation check.

Templates/TodoList-NodeJS/index.js
```diff
@@ -0,0 +1,45 @@
...
+};
+
+var methods = {
+    get: function (context, req, inputList) {
+        var itemList = [];
+
+        // Filtering the input list to only get Item and completed column
+        for (var i in inputList) {
+            itemList.push({
+                "Item": inputList[i].item, "Completed": inputList[i].completed
```
mathewc: Recommend using lower case for these JSON property names

Templates/TodoList-NodeJS/index.js
```diff
@@ -0,0 +1,45 @@
...
+            responseBody = "List Item Created";
+        }
+
+        context.res = {
+            status: statusCode,
+            body: responseBody
+        };
+    }
+    
+    // Updating an existing entity is currently supported only through C#
```
mathewc: I'd remove this comment line. When you say this is possible in C#, you mean by using the storage SDK directly? You can do that in Node as well.

Templates/TodoList-NodeJS/readme.md
```diff
@@ -0,0 +1,14 @@
+## Todo List with Node.js via HttpTrigger & Storage Table
+HttpTrigger allows you to implement CRUD implementation over Rest API. This template shows how to use a httptrigger create and read entities that resembles todo list entities from Storage Table.
```
mathewc: fix up wording

"This template implements a simple HTTP CRUD API over an Azure Storage Table, allowing todo item entities to be created and read."

Templates/TodoList-NodeJS/readme.md
```diff
@@ -0,0 +1,14 @@
...
+
+## How it works
+The http methods are mapped to CRUD operation on an Azure Storage Table binding. `GET` in this function is mapped to `Read all entities` and `POST` is mapped to `Create a new entity`. Http POST expects JSON data as shown below.
```
mathewc: operations

Templates/TodoList-NodeJS/readme.md
```diff
@@ -0,0 +1,14 @@
...
+```JSON
+{
+    "item": "todo item",
+    "completed": false
+}
+```
+
+## Limitations
+Node.js currently does not support modifying/deleting an existing entity through Table Storage binding. Hence this template is limited to Read and Create operations only.
```
mathewc: As per my comment above, I don't believe the C# binding supports updates either?

Templates/TodoList-NodeJS/function.json
```diff
@@ -0,0 +1,33 @@
...
+      "direction": "in",
+      "name": "req"
+    },
+    {
+      "type": "http",
+      "direction": "out",
+      "name": "res"
+    },
+    {
+      "type": "table",
```
mathewc: The problem here is that these input bindings are all eagerly evaluated before the method is invoked. So even for a POST, we'd be executing the query to get all the items. I don't think we want to do that. I think for something like this you'd want to do an IBinder style binding at runtime conditionally. We have a work item for enabling that in Node, but it isn't there yet.

soninaren:
Updated PR as per comments.

Templates/TodoList-NodeJS/function.json
mathewc: I made all the other comments before this one - I don't think this is a pattern we'd want to lead people to follow.

soninaren: Ideally this function would be just an empty mapper function. That calls other functions specific to each http method. This template is just a way to provide possible things to do with templates. We would have a corresponding sample in the samples repo that shows how to do this using several functions. I can may be add some comments in the code and readme.md saying that a better way to do this would be with seperate functions and point to alink in the samples repo?

mathewc: I don't think we should have a template that shows people how to do the wrong thing - in this case perform unecessary/inefficient operations. The fact is that a Function isn't suited to a general REST endpoint (either Node or C#).

soninaren: Closing the PR as per our discussion. :disappointed:

PR #197, opened 2016-05-12
Created Get and Set Entity for table storage via HttpTrigger

soninaren:


soninaren:
fixes #195

christopheranderson:
:shipit: (some changes I'd like to make, but since this is in experimental, no worries)

PR #200, opened 2016-05-14
Created Get and Set Entity for table storage via HttpTrigger for C#

soninaren:


soninaren:
fixes #196

soninaren:
@christopheranderson have updated both PR's as per our discussion.

christopheranderson:
:shipit:

Issue #186, opened 2016-05-10
BlobTrigger Run Input format must be a valid blob path


mathewc:
Currently our "Run" functionality for blob triggers is broken. The data specified in the Run box must be a valid blob path in the form container/name. If you try it you'll see the runtime emits an error to this effect.

So we should update our default sample data to lead people to the correct path. Recommend ensuring the rest of the samples work as well.

Note that the bindings themselves define this format, not the Functions runtime. Each binding chooses what format it requires for string invoke.

PR #201, opened 2016-05-16
Updated blob trigger run now text box field data

soninaren:


soninaren:
fixes #186

Templates/BlobTrigger-CSharp/sample.dat
```diff
@@ -1 +1 @@
-My sample data
\ No newline at end of file
+samples-workitems/workitem.txt
```
mathewc: Hows come the BlobTrigger-Batch template doesn't have sample data?

soninaren: Will add sample.dat to batch as well.

PR #204, opened 2016-05-18
Adding C# function for posting git hub comments

soninaren:


soninaren:
Fixes #203 

@fabiocav  can you take a look?

Templates/GitHubCommenter-CSharp/run.csx
```diff
@@ -0,0 +1,50 @@
...
+}
+
+public static async Task SendGitHubRequest(string url, string requestBody)
+{
+    using (var client = new HttpClient())
+    {
+        client.DefaultRequestHeaders.UserAgent.Add(new ProductInfoHeaderValue("username", "version"));
+
+        // Add the github access token as an app setting
+        client.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Basic", Environment.GetEnvironmentVariable("GITHUB_CREDENTIALS"));
```
fabiocav: Would probably be helpful to enhance the comment since this expects the value to be the Base64 encoded string used for basic auth.

Templates/GitHubCommenter-CSharp/run.csx
```diff
@@ -0,0 +1,50 @@
+// Please follow the link to get information on GitHub authentication   
+// https://developer.github.com/v3/oauth/
+
+using System.Net;
+using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Text;
+
+public static async void Run(dynamic payload, TraceWriter log)
```
fabiocav: Change the signature from void to Task

fabiocav:
@soninaren added a couple of comments. It would be great to use Octokit, but since this is a template, there would be issues pulling the package when it is deployed.

PR #205, opened 2016-05-18
Adding ApiHub File templates

safihamid:


Templates/SaaSFileTrigger-Batch/function.json
```diff
@@ -0,0 +1,18 @@
+{
+  "bindings": [
+    {
+      "type": "apiHubFileTrigger",
+      "name": "input",
+      "direction": "in",
+      "path": "input-path-batch/{name}",
```
mathewc: I'd simplify this from "input-path-batch" to just "input-batch". Same for output below, and all other templates.

Templates/SaaSFileTrigger-Batch/metadata.json
```diff
@@ -0,0 +1,11 @@
+{
+    "name": "SaaSFileTrigger - Batch",
+    "description": "A Batch function that will be run whenever a file is added to a SaaS provider.",
```
mathewc: What if the file is added when the function isn't running, or what if the trigger fails (e.g. due to some non user issue) before the user function is called. Is that addition lost, or does ApiHub manage determining whether a notification is actually processed successfully (i.e. retry semantics)?

Templates/SaaSFileTrigger-Batch/function.json
safihamid: It just adds to confusion. people will try creating different functions for different languages in the same function app and would not see the triggers being fired because it might have been picked up by another trigger function.
Does that make sense?

safihamid: oh sorry you mean remove the path part :-)

Templates/SaaSFileTrigger-Batch/run.bat
```diff
@@ -0,0 +1,5 @@
+@echo off
+
+type %input% >> %output%
```
mathewc: might consider using `copy %input% %output%` like the BlobTrigger-Batch sample does.

Templates/SaaSFileTrigger-Batch/run.bat
```diff
@@ -0,0 +1,5 @@
...
+
+echo bat file processed a file!
```
mathewc: nit: use the same format the other Batch samples use "Windows Batch script processed ..."

Templates/SaaSFileTrigger-Batch/metadata.json
safihamid: That's a separate issue we are tracking. https://github.com/Azure/azure-webjobs-sdk-extensions/issues/96

You may add this comment there.

Templates/SaaSFileTrigger-CSharp/run.csx
```diff
@@ -0,0 +1,8 @@
+using System;
+
+public static void Run(string input, out string output, TraceWriter log)
+{
+    log.Info($"C# processed a file!");
```
mathewc: nit: "C# SaaS trigger function processed..."

Templates/SaaSFileTrigger-NodeJS/index.js
```diff
@@ -0,0 +1,6 @@
+module.exports = function (context, input) {
+    context.log('Node.js processed a file!');
```
mathewc: nit: "Node.js SaaS trigger function processed..."

Templates/SaaSFileTrigger-Batch/metadata.json
mathewc: That issue doesn't really seem to be about what I'm asking here. We can take this one offline - I only mentioned it here because I just thought of it.

Templates/SaaSFileTrigger-Batch/function.json
mathewc: Yep, that's all :)

mathewc:
:shipit:

soninaren:
Add sample.bat to each template. The contents of sample.dat are pre-populated in the portal.

soninaren:
:+1:

PR #207, opened 2016-05-20
Fixing Bash template casing issue

mathewc:
I've verified that this fixes the template in prod.


soninaren:
:shipit:

soninaren:
we can do a quick dev to master push as well since everything else is already in master.
this will be available to customers.

mathewc:
Sure. I just committed this to dev.

PR #209, opened 2016-05-24
Adding IO bindings for apiHubTable types

Tuesdaysgreen:


Bindings/bindings.json
```diff
@@ -474,6 +474,124 @@
...
+                            "errorText": "[variables('parameterName')]"
+                        }
+                    ]
+                },
+                {
+                    "name": "dataSetName",
+                    "value": "string",
+                    "defaultValue": "default",
+                    "required": false,
+                    "label": "Data Set Name",
```
mathewc: Should DataSet be one word (here and below)?

Bindings/bindings.json
```diff
@@ -474,6 +474,124 @@
...
+                    "help": "The ID of the entity to bind to.",
+                    "validators": []
+                },
+                {
+                    "name": "connection",
+                    "value": "string",
+                    "resource": "ApiHub",
+                    "required": true,
+                    "label": "SaaS Table connection",
+                    "help": "The connection string to your SaaS Table connection",
```
mathewc: nit: "The connection string for your SaaS Table connection."

Bindings/bindings.json
```diff
@@ -474,6 +474,124 @@
             ]
         },
         {
+            "type": "apiHubTable",
+            "displayName": "SaaS Table",
```
mathewc: since this is repeated, I wonder if you can't use the "variable" syntax reference used elsewhere?

Bindings/bindings.json
```diff
@@ -474,6 +474,124 @@
...
+            "direction": "in",
+            "settings": [
+                {
+                    "name": "name",
+                    "value": "string",
+                    "defaultValue": "input",
+                    "required": true,
+                    "label": "Parameter name",
+                    "help": "This is the name of the parameter in the code for trigger to bind to.",
```
mathewc: this isn't a trigger binding

Bindings/bindings.json
```diff
@@ -474,6 +474,124 @@
...
+            "displayName": "SaaS Table",
+            "direction": "out",
+            "settings": [
+                {
+                    "name": "name",
+                    "value": "string",
+                    "defaultValue": "output",
+                    "required": true,
+                    "label": "Parameter name",
+                    "help": "This is the name of the parameter in the code for trigger to bind to.",
```
mathewc: same here as above

Bindings/bindings.json
Tuesdaysgreen: It looks like data set can be either two words or one word.  I personally feel like it should be two words so I'll leave it unless you feel strongly about it.

Bindings/bindings.json
Tuesdaysgreen: Fixed

Bindings/bindings.json
Tuesdaysgreen: I didn't even realize there was a variables syntax.  It looks like we're not using it very consistently, but I'll add it for this binding type.

Bindings/bindings.json
Tuesdaysgreen: I added a variable and fixed the string

Bindings/bindings.json
mathewc: Ok

mathewc:
:ship:

PR #210, opened 2016-05-25
Simplifying some of the standard templates

mathewc:
For most of the standard templates (e.g. queue trigger, blob trigger) the templates should be a clean starting point, just receiving the trigger input, not performing convoluted operations on additional tables, etc. I shouldn't have to configure a table connection when starting from a queue template.

I think we're struggling with defining the difference between templates that people use day to day as starting points, v.s. samples that demonstrate different capabilities. I think many of the templates we have now are samples that should be moved to the samples repo.

### DIFF
--- a/Templates/BlobTrigger-Batch/function.json
+++ b/Templates/BlobTrigger-Batch/function.json
@@ -5,14 +5,7 @@
             "name": "input",
             "type": "blobTrigger",
             "direction": "in",
-            "path": "samples-input/{name}",
-            "connection": ""
-        },
-        {
-            "type": "blob",
-            "name": "output",
-            "direction": "out",
-            "path": "samples-output/{name}",
+            "path": "samples-input",
             "connection": ""
         }
     ]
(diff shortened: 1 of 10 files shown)


mathewc:
@soninaren @christopheranderson @davidebbo

Templates/BlobTrigger-Batch/metadata.json
```diff
@@ -5,7 +5,6 @@
     "language": "Batch",
     "category": [ "Experimental" ],
     "userPrompt": [
-        "connection",
-        "path"
```
soninaren: Why remove path, we can leave this in, Since it also populated in the function.json. It should appear in the portal pre-populated? 
This is more of a subjective opinion. But leaving it in would give the user an option to change container/blob.

mathewc: That was actually a mistake - good catch :) I had removed 'path' from the other templates since not needed there. Will add back.

soninaren:
:shipit:

Templates/BlobTrigger-Batch/metadata.json
mathewc: Added back

mathewc:
Merged

PR #215, opened 2016-05-31
[NotificationHub] Add drop down to select notification platform

pragnagopa:


soninaren:
:shipit: 
Just a quick note, Selecting template in the dropdown would remove not add anything to function.json or remove the existing platform setting from function.json

PR #233, opened 2016-06-17
Adding SendGrid output binding and templates

mathewc:
See corresponding addition to the runtime [here](https://github.com/Azure/azure-webjobs-sdk-script/commit/81fbaa46cc4f773840c0b5c2c4b62f56270d711a). We won't check in this template work until after that is deployed.

Question: When using this binding, a **AzureWebJobsSendGridApiKey** app setting is required to be configured. Where is the right place to indicate that dependency to the user?

### DIFF
--- a/Bindings/bindings.json
+++ b/Bindings/bindings.json
@@ -389,7 +389,7 @@
                     "required": true,
                     "label": "Path",
                     "help": "The output file path",
-                    "validators": []
+                    "validators": [ ]
                 },
                 {
                     "name": "connection",
@@ -430,7 +430,7 @@
                     "required": true,
                     "label": "Path",
                     "help": "The input file path",
-                    "validators": []
+                    "validators": [ ]
                 },
                 {
                     "name": "connection",
@@ -471,7 +471,7 @@
                     "required": true,
                     "label": "Path",
                     "help": "The path to trigger on. This path must exist.",
-                    "validators": []
+                    "validators": [ ]
                 },
                 {
                     "name": "connection",
@@ -517,7 +517,7 @@
                     "required": false,
                     "label": "[variables('apiHubTableDataSetLabel')]",
                     "help": "[variables('apiHubTableDataSetHelp')]",
-                    "validators": []
+                    "validators": [ ]
                 },
                 {
                     "name": "tableName",
@@ -526,7 +526,7 @@
                     "required": false,
                     "label": "[variables('apiHubTableNameLabel')]",
                     "help": "[variables('apiHubTableHelp')]",
-                    "validators": []
+                    "validators": [ ]
                 },
                 {
                     "name": "entityId",
@@ -535,7 +535,7 @@
                     "required": false,
                     "label": "[variables('apiHubTableEntityLabel')]",
                     "help": "[variables('apiHubTableEntityHelp')]",
-                    "validators": []
+                    "validators": [ ]
                 },
                 {
                     "name": "connection",
@@ -576,7 +576,7 @@
                     "required": false,
                     "label": "[variables('apiHubTableDataSetLabel')]",
                     "help": "[variables('apiHubTableDataSetHelp')]",
-                    "validators": []
+                    "validators": [ ]
                 },
                 {
                     "name": "tableName",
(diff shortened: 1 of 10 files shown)


Bindings/bindings.json
```diff
@@ -389,7 +389,7 @@
                     "required": true,
                     "label": "Path",
                     "help": "The output file path",
-                    "validators": []
+                    "validators": [ ]
```
mathewc: I opened this in VS and VS is formatting all this :)

Bindings/bindings.json
```diff
@@ -1359,6 +1359,57 @@
...
+                {
+                    "name": "to",
+                    "value": "string",
+                    "defaultValue": "",
+                    "required": false,
+                    "label": "To address",
+                    "help": "To address that the message should be sent to.",
+                    "validators": [
+                        {
+                            "expression": "TODO",
```
mathewc: Do we have a standard regex for email addresses that we use somewhere in the portal?

soninaren: got it from the portal.azure.com aad portal extension
^(|[a-zA-Z0-9._-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,4})$

Bindings/bindings.json
```diff
@@ -1359,6 +1359,57 @@
                     "help": "Choose platform to send native notifcations. Select Template to send template notification."
                 }
             ]
+        },
+        {
+            "type": "sendGrid",
```
soninaren: This type needs to be added to portal

Bindings/bindings.json
```diff
@@ -1359,6 +1359,57 @@
...
+                        }
+                    ]
+                },
+                {
+                    "name": "subject",
+                    "value": "string",
+                    "defaultValue": "",
+                    "required": false,
+                    "label": "Message Subject",
+                    "help": "The Subject line to use for the message."
```
soninaren: Need regex validators for this and the text field as well. (if possible)

Bindings/bindings.json
```diff
@@ -1359,6 +1359,57 @@
...
+            "displayName": "SendGrid",
+            "direction": "out",
+            "settings": [
+                {
+                    "name": "to",
+                    "value": "string",
+                    "defaultValue": "",
+                    "required": false,
+                    "label": "To address",
+                    "help": "To address that the message should be sent to.",
```
soninaren: Nit: The help messages in the portal are formatted in the following manner. May be should change to keep them consistent.
"This is the email address the message should be sent to."

Bindings/bindings.json
soninaren: Ahh the regex is purely for email address

mathewc: I'll just leave the regex out for now. We can add one later.

Bindings/bindings.json
mathewc: Email Text and Subject line are pretty open ended. I don't think we need a regex here.

Templates/SendGrid-CSharp/metadata.json
```diff
@@ -0,0 +1,17 @@
+{
+    "defaultFunctionName": "SendGridCSharp",
+    "description": "A C# function that sends a SendGrid confirmation mail whenever a new order is added to a specified Azure Storage queue",
+    "name": "SendGrid - C#",
+    "language": "C#",
+    "category": [
+        "Samples"
```
mathewc: Moved these to the samples category

Bindings/bindings.json
mathewc: That has to happen AFTER these templates are checked in, right? I'll log an issue for that.

soninaren:
:shipit:

soninaren:
Generally i mention about adding app setting in the code file via comment.

Bindings/bindings.json
mathewc: I've updated them to use that form "This is the..."

mathewc:
See last revision - I've added comment blocks to the templates indicating required/optional configuration.

Issue #238, opened 2016-06-20
Fix the case for documentdb binding type


soninaren:
We need to redo this [commit](https://github.com/Azure/azure-webjobs-sdk-templates/commit/48fd7048f9b50da041d0e28e2950070218029c27) once this https://github.com/projectkudu/AzureFunctionsPortal/issues/411 issue is fixed

mathewc:
Naren, can you reaply this now that the portal fix is in? Does the release need to be staged?

soninaren:
need to redeploy portal before merging to master, can be tested in the dev portal.

soninaren:
#253 fixed

PR #239, opened 2016-06-21
reapplying the case change for document db binding

soninaren:


mathewc:
:ship:

mathewc:
The portal needs to be deployed before this, right? Also, can you guys verify in stage before releasing?

soninaren:
yeah, we need to deploy portal. We can verify on dev portal as soon as this commit is merged in

soninaren:
fixes #238

PR #241, opened 2016-06-22
Adding Consumer Group to the EventHub binding

mathewc:
Support for EH consumer group is going out in the upcoming release, so we need to expose this.

### DIFF
--- a/Bindings/bindings.json
+++ b/Bindings/bindings.json
@@ -89,6 +89,20 @@
                     "label": "Event Hub connection",
                     "help": "This is the connection string for your event hub. Make sure that it has Receive permissions.",
                     "placeholder": "[variables('selectConnection')]"
+                },
+                {
+                    "name": "consumerGroup",
+                    "value": "string",
+                    "defaultValue": "$Default",
+                    "required": false,
+                    "label": "Event Hub consumer group",
+                    "help": "The name of the Event Hub consumer group to receive events from.",
+                    "validators": [
+                        {
+                            "expression": "^([a-z0-9]$|^[a-z0-9][a-z0-9-_.]{0,48}[a-z0-9])|(\\$Default)$",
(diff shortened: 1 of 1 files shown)


Bindings/bindings.json
```diff
@@ -89,6 +89,14 @@
...
                     "help": "This is the connection string for your event hub. Make sure that it has Receive permissions.",
                     "placeholder": "[variables('selectConnection')]"
+                },
+                {
+                    "name": "consumerGroup",
+                    "value": "string",
+                    "defaultValue": "$Default",
+                    "required": false,
+                    "label": "Event Hub consumer group",
+                    "help": "The name of the Event Hub consumer group to receive events from."
```
mathewc: Need to find out what if any validator can be applied. Until we know for sure, we won't have one.

Templates/EventHubTrigger-CSharp/function.json
```diff
@@ -6,7 +6,8 @@
             "name": "myEventHubMessage",
             "direction": "in",
             "path": "samples-workitems",
-            "connection": ""
+            "connection": "",
+            "consumerGroup": "$Default"
```
mathewc: The runtime will default this if not specified. Should I put these default values here or remove them and let the template take care of it?

Bindings/bindings.json
mathewc: Got confirmation from the EH team that we can just use the Storage naming rules, and whitelist $Default

Templates/EventHubTrigger-CSharp/function.json
soninaren: Typically non mandatory fields are not added to templates, if the customer needs that explicitly they can add the value from integration tab, which then will just add the setting to function.json

soninaren:
:shipit:

Templates/EventHubTrigger-CSharp/function.json
mathewc: Ok, I'll remove these defaults.

PR #244, opened 2016-06-23
New Node template for generating an Azure Storage SAS token

lindydonna:
Code has been reviewed by Dale.

### DIFF
--- a/Templates/SasToken-Node/function.json
+++ b/Templates/SasToken-Node/function.json
@@ -1,0 +1,17 @@
+{
+  "bindings": [
+    {
+      "authLevel": "function",
+      "type": "httpTrigger",
+      "direction": "in",
+      "name": "req",
+      "methods": [ "post" ]
+    },
+    {
+      "type": "http",
+      "direction": "out",
+      "name": "res"
+    }
+  ],
+  "disabled": false
+}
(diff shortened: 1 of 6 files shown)


Templates/SasToken-Node/metadata.json
```diff
@@ -0,0 +1,13 @@
+{
+    "defaultFunctionName": "SasTokenNodeJS",
+    "description": "A Node.js function that generates a SAS token for Azure Storage for a given container and blob name",
+    "name": "SAS Token Generator - Node",
+    "language": "JavaScript",
+    "category": [
+        "Samples",
+        "API & Webhooks"
```
mathewc: This should only be in Samples, not in the core API & WebHooks, since it is a pretty specialized sample scenario. For all our other existing samples, I don't think we cross categorize any of them like this?

lindydonna:
@mathewc I've changed the category to just Samples and added a "methods" param to function.json. Could you sign-off?

Templates/SasToken-Node/function.json
```diff
@@ -0,0 +1,18 @@
+{
+  "bindings": [
+    {
+      "authLevel": "function",
+      "webHookType": "",
```
soninaren: No need for the webHookType setting

mathewc: nit: remove this uneccessary webHookType (unless you put it in there intentionally for a reason). Our default http templates don't include this.

mathewc:
One minor comment, then :ship:

lindydonna:
@mathewc @soninaren Thanks for the feedback, the PR is now updated and commits squashed.

Issue #129, opened 2016-04-05
Add Try-Functions metadata


paulbatum:
We need metadata for both bindings and templates that indicates whether they are applicable in try functions mode (because in this mode you have access to a limited number of resource types e.g. you have storage but not NH/ServiceBus/DocDB/etc)

alrod:
@soninaren, are you going to start working on this?

soninaren:
I have not received any word about backend work done on this one. At the moment it is in backlog.

soninaren:
#271

PR #264, opened 2016-07-15
Add disabledInTryMode to support try mode

fashaikh:
Updated bindings and templates to include disabledInTryMode


mathewc:
Can you provide more details on this? I haven't heard anything about this.

soninaren:
Right now the only resource you can provision in try mode is Storage. So only templates & bindings with webhooks, timers and storage are enabled in try mode. Everything else would be disabled (Greyed out).

Templates/FaceLocator-NodeJS/metadata.json
```diff
@@ -11,5 +11,6 @@
         "path",
         "connection",
         "tableName"
-    ]
+    ],
+    "disabledInTryMode": false
```
mathewc: Wondering if we shouldn't invert the flag to be "tryModeEnabled" and only explicitly set that flag to true for the ones we support in try mode. That way it's opt in, and all templates are disabled in try mode unless we explicitly set this. That makes it less likely that someone accidentally adds a template that shows up but doesn't work in try mode.

soninaren: That makes sense
@fashaikh can you inverse the flag. This way we would also have a smaller change.

fashaikh: Okay. Will make the change

fashaikh:
Updated naming. Referencing #129

Templates/default/Bindings/bindings.json
```diff
@@ -0,0 +1,1382 @@
+{
```
soninaren: This file is not needed.

Templates/ServiceBusTopicTrigger-CSharp/metadata.json
```diff
@@ -12,5 +12,6 @@
         "accessRights",
         "topicName",
         "subscriptionName"
-    ]
+    ],
+    "disabledInTryMode": true
```
soninaren: change this to enabledInTryMode

Templates/Resources/Resources.resx
```diff
@@ -0,0 +1,276 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<root>
```
soninaren: There are a few bugs in the strings, please check this commit in my fork and apply the same,
https://github.com/soninaren/azure-webjobs-sdk-templates/commit/dd947cfb3a9db36ff4762f1ba5a4285728d4a860

soninaren:
creating a new PR because of conflicts and localization changes

soninaren:
#271

PR #267, opened 2016-07-22
Add FTP and SFTP to ApiHubFile triggers and remove Azure Blob 

safihamid:
Add FTP and SFTP to ApiHubFile triggers and remove Azure Blob from list of triggers


emilcicos:
:shipit:

PR #275, opened 2016-07-27
Adding the regex for comsumer group

soninaren:


soninaren:
blocked on https://github.com/projectkudu/AzureFunctionsPortal/issues/430

Issue #281, opened 2016-08-08
Add Documentation for Input binding, output binding and triggers


soninaren:

soninaren:
I have added place holder files and settings in binding.json for in PR #284

soninaren:
#288 fixed this

PR #284, opened 2016-08-11
Added documentation setting for all bindingTypes

soninaren:


soninaren:
Addressed #281

soninaren:
@alrod can you take a look

PR #287, opened 2016-08-18
Template for HTTPGet(CRUD)-PHP

Eun:
However I don't know howto create the translation for `Resources/qps-ploc`


soninaren:
Thanks for the contribution @Eun. Will look through this and merge it in.

PR #291, opened 2016-08-27
Updating F# templates to work with new model

fabiocav:
Updating existing F# samples to work with the new F# language model.

**NOTE**: This will only work with the next release (runtime version 0.5.x +) and should not be pushed before that is available.

### DIFF
--- a/Templates/QueueTrigger-FSharp/readme.md
+++ b/Templates/QueueTrigger-FSharp/readme.md
@@ -1,0 +1,11 @@
+# QueueTrigger - F<span>#</span>
+
+The `QueueTrigger` makes it incredibly easy to react to new Queues inside of Azure Queue Storage. This sample demonstrates a simple use case of processing data from a given Queue using F#.
+
+## How it works
+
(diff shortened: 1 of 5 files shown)


Templates/QueueTrigger-FSharp/readme.md
```diff
@@ -0,0 +1,11 @@
+# QueueTrigger - F<span>#</span>
+
+The `QueueTrigger` makes it incredibly easy to react to new Queues inside of Azure Queue Storage. This sample demonstrates a simple use case of processing data from a given Queue using F#.
+
+## How it works
+
+For a `QueueTrigger` to work, you provide a path which dictates where the queue messages are located inside your container.
```
mathewc: I think the C# source you copied this from is also incorrect - it also talks about "path" and "container" which was likely copied from blob trigger. You don't specify a "path", you specify a queue name.

soninaren:
Looks good.

fabiocav:
Updated the documentation (including C#), since, as Mathew pointed out, it was incorrect.

Issue #51, opened 2016-03-21
Allow bindings to be categorized (specifically into an 'experimental' category)


brettsam:
Some bindings won't behave the same when used in node vs when used in C#. It'd be nice if we could restrict a binding to supported languages until they're updated, tested, and verified.

mathewc:
Having a setting like that is dangerous I think - it could lead to us saying some bindings don't work across languages, which we don't want. What binding are you referring to specifically that doesn't work in Node?

Can't we just use the existing "experimental" category to classify templates that aren't yet working?

brettsam:
This issue: https://github.com/Azure/azure-webjobs-sdk-script/issues/107.

This means that C# `in` bindings will update the changes to the object when it exits (it's effectively `inout` but C# doesn't care about what you specify on the binding). But in node it will only act as `in` and any changes you make in the function do not get sent along.

My guess is we'll bump into other things like this in the future and it's nice to be able to restrict behavior. In this example, I'd rather be able to say -- if you're using `in`, then you need C# right now.

I don't think you can classify bindings as experimental -- that's just for templates.

fabiocav:
I'm with @mathewc on this one. I'd rather see bindings placed in an "experimental" category than showing up as production ready if there are discrepancies of different support depending on the features you're using.

brettsam:
Okay -- I've updated the title.

mathewc:
I don't think we should categorize bindings in this way. We have bugs and we'll fix them, but the goal is that any binding we release templates for works across languages, with a few templates for each. There will be some differences across languages (there are other examples outside of this in/out issue). That's ok and expected I think.

alrod:
Portal change is done. example of using:
        {
            "type": "eventHubTrigger",
            "displayName": "Azure Event Hub",
            "direction": "trigger",
            "defaultParameterName": "myEventHubMessage",
            "parameterNamePrompt": "Event parameter name",
            "settings": [
                {
                    "name": "path",
                    "value": "string",
                    "defaultValue": "myEventHub",
                    "required": true,
                    "label": "Event Hub name",
                    "help": "This is the name of the Event Hub you would like to trigger off of.",
                    "validators": [
                        {
                            "expression": "^[A-Za-z0-9]$|^[A-Za-z0-9][\\w-\\.\\/]*[A-Za-z0-9]$",
                            "errorText": "Invalid event hub name"
                        }
                    ]
                },
                {
                    "name": "connection",
[...]

PR #292, opened 2016-08-27
Adding Twilio SMS binding

fabiocav:


Resources/Resources.resx
```diff
@@ -966,6 +966,45 @@
   <data name="timerTrigger_schedule_errorText" xml:space="preserve">
     <value>Invalid Cron Expression. Please consult the &lt;a target='_blank' href='https://azure.microsoft.com/en-us/documentation/articles/functions-bindings-timer/'&gt;documentation&lt;/a&gt; to learn more.</value>
   </data>
+    <data name="twilioSms_accountsid_help" xml:space="preserve">
```
fabiocav: Only updated the language neutral resource file. I'm assuming the others are generated from this and we can properly fallback if the resource is not found in the culture specific resource files.

fabiocav:
Also regenerated the `Resources.Designer.cs` file. Doesn't seem like it was required, since it was missing a lot of resources, but this brings it up to date.

mathewc:
Looks OK to me. Looks like a lot of stuff has changed in templates (loc, etc.) so those guys should take a look as well.

Resources/Resources.resx
```diff
@@ -966,6 +966,45 @@
...
+    <value>This is your Twilio account SID</value>
+  </data>
+  <data name="twilioSms_accountsid_label" xml:space="preserve">
+    <value>Account SID</value>
+  </data>
+  <data name="twilioSms_authtoken_help" xml:space="preserve">
+    <value>This is your Twilio Auth Token</value>
```
soninaren: Nit: Missing "." at the end of line.

soninaren:
We can put this behind filter like this and merge it in.

```JSON
"filters": [
                "showHiddenBinding"                
              ]
```
refer to this for more details
https://github.com/Azure/azure-webjobs-sdk-templates/issues/51

soninaren:
Looks good.

fabiocav:
Out of curiosity, what is the filter for?

soninaren:
So that we can hide a binding in production until it is ready for release.

fabiocav:
Technically, there's no reason to hide that binding. I'm assuming what we add here will be available in the next release (next tag), correct?

mathewc:
Yeah, I don't think this needs to be hidden either. It just needs to be tagged along with the other changes like F# template updates, etc.

soninaren:
Yeah we should not hide it.

PR #304, opened 2016-09-20
Updating parameter name validation regex

soninaren:


soninaren:
@fabiocav can you take a look?

Bindings/bindings.json
```diff
@@ -35,7 +35,7 @@
                     "help": "$timerTrigger_name_help",
                     "validators": [
                         {
-                            "expression": "^[a-zA-Z_$][a-zA-Z_$0-9]*$",
+                            "expression": "^[a-zA-Z][a-zA-Z0-9]{0,127}$",
```
mathewc: Allow for 1

mathewc:
:shi"

Bindings/bindings.json
mathewc: Confirmed with Fabio that this is correct

PR #305, opened 2016-09-20
Changing text to remove the trailing # in C#

fashaikh:
marked ignores the trailing `#`. I tried escaping with `\` but that didn't work. 
A trailing `&nbsp;` is an option , where we get to keep the text as is, but I don't know how it'll look like in parsers.

### DIFF
--- a/Documentation/tableIn.md
+++ b/Documentation/tableIn.md
@@ -64,7 +64,7 @@ module.exports = function (context, myQueueItem) {
 };
 ```
 
-#### Storage tables example: C# example tha reads multiple table entities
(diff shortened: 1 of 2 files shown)


mathewc:
What library can't handle trailing #'s? This seems like a fragile place to be - we're going to keep accidentally adding trailing #'s. What breaks when that happens?

fashaikh:
[This](https://github.com/projectkudu/AzureFunctionsPortal/blob/dev/AzureFunctions/Default.cshtml#L38) library seems to be parsing the markdown into html. 
 For trailing # where the markdown begins with a # it seems to ignore the trailing # and displays C# as C

image

PR #308, opened 2016-09-26
updating SendGrid bindings and templates with latest changes

brettsam:
See https://github.com/Azure/azure-webjobs-sdk-extensions/pull/152 for a description of the underlying changes in Extensions.

### DIFF
--- a/Bindings/bindings.json
+++ b/Bindings/bindings.json
@@ -1474,6 +1474,14 @@
                     ]
                 },
                 {
+                    "name": "apiKey",
+                    "value": "string",
+                    "defaultValue": "SendGridApiKey",
+                    "required": true,
+                    "label": "$sendGrid_apiKey_label",
+                    "help": "$sendGrid_apiKey_help"
+                },
+                {
                     "name": "to",
                     "value": "string",
                     "defaultValue": "",
--- a/Templates/SendGrid-CSharp/function.json
+++ b/Templates/SendGrid-CSharp/function.json
@@ -10,6 +10,7 @@
             "type": "sendGrid",
             "name": "message",
             "direction": "out",
+            "apiKey":  "SendGridApiKey",
             "from": "Azure Functions <samples@functions.com>",
             "to": "{CustomerEmail}"
         }
--- a/Templates/SendGrid-NodeJS/function.json
+++ b/Templates/SendGrid-NodeJS/function.json
@@ -10,6 +10,7 @@
             "type": "sendGrid",
             "name": "message",
             "direction": "out",
+            "apiKey": "SendGridApiKey",
             "from": "Azure Functions <samples@functions.com>",
             "to": "{customerEmail}"
         }
--- a/Resources/Resources.resx
+++ b/Resources/Resources.resx
@@ -1011,4 +1011,10 @@
   <data name="documentDB_displayName" xml:space="preserve">
     <value>Azure DocumentDB Document</value>
   </data>
+  <data name="sendGrid_apiKey_help" xml:space="preserve">
+    <value>This is your SendGrid API Key</value>
+  </data>
+  <data name="sendGrid_apiKey_label" xml:space="preserve">
+    <value>SendGrid API Key</value>
+  </data>
 </root>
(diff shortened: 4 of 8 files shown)


Bindings/bindings.json
```diff
@@ -1474,6 +1474,14 @@
                     ]
                 },
                 {
+                    "name": "apiKey",
+                    "value": "string",
+                    "defaultValue": "SendGridApiKey",
```
mathewc: Our existing setting name is AzureWebJobsSendGridApiKey, and that is what the extension is looking for, if not set in the binding, right? Should we change the name, or keep the existing name? For existing users who have this setting already, it'd be nice if the new template just worked.

mathewc:
:ship:

Bindings/bindings.json
brettsam: I did think about that, but ultimately I based this on the Twilio setup, which gives shorter setting name like this.  I was trying to pick a short, easy-to-remember value here to optimize for people manually entering this into App Settings since this won't have a picker. I didn't see any bindings that have default values of the `AzureWebJobs*` variety anywhere else, either -- granted a lot of those have pickers, though, so it doesn't apply.

mathewc: Ok, I can see value in keeping the better name you have anyways. We'll just have to include details in the release notes.

mathewc: Maybe it's the case that existing apps with the AzureWebJobsSendGridApiKey will continue to work if they don't specify the setting in the binding? So to migrate, people can just fix their function code, but won't have to add a new app setting.

mathewc:
@soninaren This will be a breaking change to the templates for SendGrid, so we'll have to stage this in the next breaking change release (the upcoming 0.7 runtime release).

Bindings/bindings.json
brettsam: Yes, existing app settings will continue to work -- if we don't see an attribute, it'll flow up to the Environment variable that they already have. You're right they'll need to go through some pain to get to the new SendGrid model, though.

brettsam:
Hey @soninaren does this all look good? Wanted to get your 2 cents before getting it in.

Templates/SendGrid-CSharp/function.json
```diff
@@ -10,6 +10,7 @@
             "type": "sendGrid",
             "name": "message",
             "direction": "out",
+            "apiKey":  "SendGridApiKey",
             "from": "Azure Functions <samples@functions.com>",
```
soninaren: Update the metadata.json for the template to prompt for this setting as well.
https://github.com/Azure/azure-webjobs-sdk-templates/blob/dev/Templates/SendGrid-CSharp/metadata.json#L10

Templates/SendGrid-NodeJS/function.json
```diff
@@ -10,6 +10,7 @@
             "type": "sendGrid",
             "name": "message",
             "direction": "out",
+            "apiKey": "SendGridApiKey",
             "from": "Azure Functions <samples@functions.com>",
```
soninaren: Update the metadata.json for the template to prompt for this setting as well.
https://github.com/Azure/azure-webjobs-sdk-templates/blob/dev/Templates/SendGrid-NodeJS/metadata.json#L11

Templates/SendGrid-CSharp/function.json
brettsam: Wait, these `userPrompt`s shouldn't have `connection` and `queueName` for SendGrid, right? https://github.com/Azure/azure-webjobs-sdk-templates/blob/dev/Templates/SendGrid-CSharp/metadata.json#L11-L12. Maybe that was a copy/paste error from a previous release?

brettsam: Note: I removed them in my latest commit. If they're needed, I can put them back, but I don't think they are.

soninaren:
:shipit:

Templates/SendGrid-CSharp/function.json
soninaren: Yes, they are definitely not needed.

PR #311, opened 2016-09-27
Display storage account information in documentation

fashaikh:
Needs to go in together with this PR : https://github.com/projectkudu/AzureFunctionsPortal/pull/592/

### DIFF
--- a/Documentation/blobIn.md
+++ b/Documentation/blobIn.md
@@ -1,3 +1,12 @@
+#### Connecting to your Storage Account
+
+* Download Storage explorer from here: http://storageexplorer.com/ 
+* Connect using these credentials:
+* Account Name: `{{accountName}}`
+* Account Key: `{{accountKey}}`
+
+You can now view the blobs, queues and tables associated with this storage binding.
+
 #### Settings for a storage blob input binding
 
 - `name` : The variable name used in function code for the blob . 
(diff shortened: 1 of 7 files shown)


Documentation/blobIn.md
```diff
@@ -1,3 +1,12 @@
+#### Connecting to your Storage Account
+
+* Download Storage explorer from here: http://storageexplorer.com/ 
+* Connect using these credentials:
+* Account Name: `{{accountName}}`
+* Account Key: `{{accountKey}}`
```
fashaikh: Hmm okay. Will experiment with hiding it

fashaikh: BTW I noticed we do display github secret in plain text
image . Still dont have a good answer on hiding this with markdown. Working on it

mathewc: Github secrets, WebHook/Http security codes, etc. we do show in plain text you're right. I agree that that is a slight discrepancy. In my mind showing the full storage credentials seems worse though.

fashaikh: I'll close this PR. The new change to hide secrets by default will directly be made in the AzureFunctionsPortal, and wont need any changes here.
Adding custom html on the fly is apparently not supported by angular.

PR #327, opened 2016-10-22
Updating Twilio binding settings descriptions

fabiocav:
Updating Twilio binding settings descriptions to make it clear that app settings should be used for Account SID and Auth Token


mathewc:
All this duplication of strings across res files is quite unfortunate.

:ship:

soninaren:
You actually don't need to put strings in all the resource files. You just fill in the English. Rest we get from a build location.

PR #333, opened 2016-10-26
Updating HTTP Route instructions

fabiocav:


paulbatum:
:shipit:  @soninaren FYI

PR #335, opened 2016-10-28
Updating auth level instructions to match the new portal experience

fabiocav:


fabiocav:
@soninaren this would need to be localized. Can you make sure the localization flow happens and we get updated localized resources?

soninaren:
sure

soninaren:
:shipit:

PR #344, opened 2016-11-08
Replacing NodeJS text with Javascript

soninaren:


lindydonna:
Overall, looks good. To reduce risk, I'd recommend backing out the changes to queue names in function.json. It's fine to have "node" there.

PR #355, opened 2016-11-14
Removing $return for httpTriggers for 0.9

soninaren:


mathewc:
You can leave the $return for Node - it works. The issue was only for C#.

PR #356, opened 2016-11-14
Update ImageResizer sample, remove System.Drawing

lindydonna:
Since automatic NuGet restore now works, remove the commented out code that used System.Drawing. Also cleaned up "using" statements that weren't needed.


soninaren:
👍 looks good

PR #361, opened 2016-11-19
Updating the templates to use $return

soninaren:


soninaren:
@mathewc  & @fabiocav  can you take a look?

PR #364, opened 2016-11-21
Remove default value for Notification Hub name

pragnagopa:
Notification Hub name is a required value and customers might miss configuring this value if there is a default value.


soninaren:
:shipit:

Issue #370, opened 2016-12-02
Fix the regex to allow named parameters for blob input and output path setting.


soninaren:
The regex right now is more restrictive. Need to go over all the regex to validate that they allow named parameters

PR #377, opened 2016-12-12
Making regex more permissive for named parameters

soninaren:


soninaren:
Fixes https://github.com/Azure/azure-webjobs-sdk-templates/issues/370

PR #392, opened 2017-01-14
Proposed: Information for batch files

MisterJames:
Sorry for the flyby PR, but I needed to work with queue bindings for batch files and couldn't find any documentation. After some experimentation I was able to figure out how they worked and thought this might be an okay place to share. 

Cheers!

### DIFF
--- a/Documentation/queueOut.md
+++ b/Documentation/queueOut.md
@@ -68,3 +68,32 @@ module.exports = function(context, myQueueItem) {
     context.done();
 }
 ```
+
+#### Batch File Bindings for Storage Queue
+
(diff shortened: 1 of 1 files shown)


soninaren:
@lindydonna @christopheranderson  can you take a look?

lindydonna:
@MisterJames thanks for the PR!

@soninaren It looks good to me in terms of the examples. However, I'm not an expert on how the batch binding config works, perhaps @mathewc should take a look?

christopheranderson:
We can ship this as-is. We have a larger work item to document the "scripting" experience for all bindings/etc. While we're doing that, we'll likely touch this doc again.

PR #393, opened 2017-01-16
Changed if..then to isNull and pattern match.  More idomatic F#

jamessdixon:


Templates/HttpPOST(CRUD)-FSharp/run.fsx
```diff
@@ -18,17 +18,14 @@ let Run(req: HttpRequestMessage, outTable: ICollector<Person>, log: TraceWriter)
...
             let person = Person()
             person.PartitionKey <- "Functions"
             person.RowKey <- Guid.NewGuid().ToString()
             person.Name <- name
             outTable.Add(person)
-
             return req.CreateResponse(HttpStatusCode.Created)
-        else
-            return req.CreateResponse(HttpStatusCode.BadRequest,
-                "Please pass a name in the request body")
```
soninaren: The behavior needs to consistent with other HttpPOST(CRUD) templates, should return a 400 when name is not passed in the request body.

PR #395, opened 2017-01-19
F sharp templates

jamessdixon:
swtiched out some if.. then for pattern matches.  Changed 2 loops to a Seq.iter

### DIFF
--- a/Templates.sln
+++ b/Templates.sln
@@ -656,7 +656,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "TimerTrigger-FSharp", "Time
 EndProject
 Global
 	GlobalSection(SolutionProperties) = preSolution
-		HideSolutionJavaScript = FALSE
+		HideSolutionNode = FALSE
 	EndGlobalSection
 	GlobalSection(NestedProjects) = preSolution
 		{2D537292-B604-40A0-932D-647124A39216} = {9131FAC9-F834-4D69-A9F3-0057D7BB5670}
--- a/Templates/ExternalTable-FSharp/run.fsx
+++ b/Templates/ExternalTable-FSharp/run.fsx
@@ -20,8 +20,8 @@ let Run(req: HttpRequestMessage, input: ITable<Contact>, log: TraceWriter) =
                 input.ListEntitiesAsync(continuationToken = cont)
                 |> Async.AwaitTask
             
-            for item in segment.Items do
-                log.Info(sprintf "%s %s" item.FirstName item.LastName)
+            segment.Items 
+            |> Seq.iter(fun i ->  log.Info(sprintf "%s %s" item.FirstName item.LastName))
 
             match segment.ContinuationToken with
             | null -> return ()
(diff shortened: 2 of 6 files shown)


soninaren:
@sylvanc can you take a look?

Templates.sln
```diff
@@ -656,7 +656,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "TimerTrigger-FSharp", "Time
 EndProject
 Global
 	GlobalSection(SolutionProperties) = preSolution
-		HideSolutionJavaScript = FALSE
```
christopheranderson: Why was this changed?

jamessdixon: Magic of Visual Studio?  That was not me (though hiding javascript is usually not a bad idea).  Do you want me to fix and resubmit?

sylvanc:
LGTM :)

dsyme:
These changes are not necessary, though thanks for proposing them

* while some F# programmers like to pattern match on single boolean true/false values, i'ts not particularly idiomatic to do that - indeed I suppose if/then/else would normally be seen as marginally more idiomatic, 

* For Seq.iter, much the same applies: some F# programmers eschew the use of "for". However there can also be good reasons to use it, notably the way it combines with computation expressions and comprehensions.

cheers and thanks
don

jamessdixon:
Who is this Don Syme and what does he know about F#?  Granted the changes aren't necessary, I have found that syntax like that does help when transitioning people from imperative constructs into a more functional style.  Since these templates are going into Visual Studio and a majority of VS Users are C# devs, it seemed like a easy point of education.

dsyme:
@jamessdixon :) I understand - yes, it's tricky balance between familiarity (more like C#) and teaching (use more F#-specific constructs). 

However for these templates I think the emphasis shouldn't be on teaching F#, just on landing people with an Azure Function that does the job for them

jamessdixon:
I was going to add some more templates.  Can I use that style or do you want consistency in the code base?

dsyme:
@jamessdixon Consistency please (unless that prevents you adding the templates :) :) :) )

jamessdixon:
The empty template in Visual Studio right now has an unneeded semi colon:

let Run(log: TraceWriter ) =
    log.Verbose("F# function executed.")**;**

Want a pull request for that?

soninaren:
Visual studio is using older version F# templates, All empty templates have been removed from the repo. Also FSharpTemplates branch is no longer in use, It was created to temporarily hold some F# Templates. Please use send all future PR's against dev branch. :)

jamessdixon:
OK - will do
When will these templates be pushing into VS?  Should I ask the VS team?

soninaren:
VS team controls that timeline, I will ask them and let you know.

jamessdixon:
Thanks- I can also ask on the MVP thread if you don't get a fast answer

christopheranderson:
Summoning @lindydonna who has the answer on VS stuff. 

FYI - our cli uses these templates when you run locally and it pulls from GitHub directly.

lindydonna:
@jamessdixon We don't have a concrete timeline on when there will be a refresh of the Azure Functions Visual Studio tools. Currently it uses a snapshot of the templates feed from 0.6. We're changing up how the tooling works, which is why I don't have anything more concrete. :)

soninaren:
From the discussion above it looks like we might not be merging the changes, would it be okay if we close this PR?

jamessdixon:
Sure - though I think Don is making a mistake.  Let the records show that I shook my fist at him.  Thanks!

soninaren:
:)

Issue #397, opened 2017-01-24
Support Service Bus topics names with hyphens


christopheranderson:
When creating a new function triggered by service bus topic, the validation on topic name field refuses hyphenated, which are actually valid for topic names.

From: https://feedback.azure.com/forums/355860/suggestions/17819737

soninaren:
fixed in #398

PR #398, opened 2017-01-24
Fixing the topic name regex

soninaren:


soninaren:
fix for https://github.com/Azure/azure-webjobs-sdk-templates/issues/397

PR #405, opened 2017-02-02
Update Event Hubs and Service Bus validation

anthonychu:
Event Hubs and Service Bus entities (event hubs, consumer groups, topics, subscriptions, queues) have the same validation rules. This PR fixes some issues with the validation and uses the same regex for Event Hubs and Service Bus.

- Upper case letters are allowed
- 1 and 2 character entity names are allowed (currently Service Bus validation does not allow this)

Corresponding validation messages (and translations) may need to be updated.

### DIFF
--- a/Bindings/bindings.json
+++ b/Bindings/bindings.json
@@ -86,7 +86,7 @@
                     "help": "$eventHubTrigger_path_help",
                     "validators": [
                         {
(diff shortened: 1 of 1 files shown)


soninaren:
This looks good, I am going to update the corresponding messages

PR #406, opened 2017-02-09
Remove node http binding return default

mamaso:
resolves https://github.com/Azure/azure-webjobs-sdk-script/issues/1182

### DIFF
--- a/Templates/HttpTrigger-JavaScript/index.js
+++ b/Templates/HttpTrigger-JavaScript/index.js
@@ -2,16 +2,16 @@
     context.log('JavaScript HTTP trigger function processed a request.');
 
     if (req.query.name || (req.body && req.body.name)) {
-        res = {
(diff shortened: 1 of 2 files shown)


Templates/HttpTrigger-JavaScript/index.js
```diff
@@ -2,16 +2,16 @@
...
             body: "Hello " + (req.query.name || req.body.name)
         };
+        context.done();
     }
     else {
-        res = {
-            status: 400,
-            body: "Please pass a name on the query string or in the request body"
-        };
+        // express-like response api
```
mathewc: Let's not mix these two different styles in the same template. I get what you're trying to achieve, but I'd like to keep the templates consistent.

mamaso: Fair enough 👍

Issue #409, opened 2017-02-15
Errors in Blob path UI and Regex Validation


lindydonna:
_From @mathewc on February 11, 2017 17:42_

2 issues. First, our blob path regex is overly restrictive. Second, the actual help text isn't resolving. Path expressions can contain an arbitrary number of path segements, and each supports {} binding expressions. Note also that %% expressions are also supported (binding from app settings, e.g. %MY_CONTAINER%/{Foo}/{Bar}). We should review all our regexes and make sure they're correct in this regard.

![image](https://cloud.githubusercontent.com/assets/1056548/22856012/54404a50-f03e-11e6-9610-eeb4973a5cc4.png)

_Copied from original issue: projectkudu/AzureFunctionsPortal#911_

lindydonna:
_From @mathewc on February 11, 2017 17:45_

@soninaren

soninaren:
Fixing the regex so that it accept the following inputs, Fixing regex for storage bindings, opening a new issue to review all other regexs
```
con/a
{container}/{a}
%abc%/{a}/{ab}/a
%abc%/%a%/{abc}/a
{Container}/{first}
con/{a}
```

Also for blob trigger it accept the following as well
```
con
{Container}
{container}
%abc%
```

PR #425, opened 2017-04-11
fixing blob path regex

soninaren:
Fixing the regex so that it accept the following inputs, Fixing regex for storage bindings, opening a new issue to review all other regexs
```
con/a
{container}/{a}
%abc%/{a}/{ab}/a
%abc%/%a%/{abc}/a
{Container}/{first}
con/{a}
```

Also for blob trigger it accept the following as well
```
con
{Container}
{container}
%abc%
```


soninaren:
#409

PR #430, opened 2017-04-12
Change node event hub cardinality

### LINKED ISSUE #429: Switch Node.JS event hubs template over to use batching
This is a more targeted version of: #293 

Specifically @tohling has seen a few customers be more successful with batching for JavaScript functions (see https://github.com/Azure/azure-webjobs-sdk-script/issues/1386#issuecomment-293372407). More investigation is required to understand exactly what factors are at play but in the meantime, users will be more successful with a batch based template.

mamaso:
@soninaren gonna need you on this one, we need to discuss what lights up in the portal / how much of that we want to do :)

resolves #429

### DIFF
--- a/Bindings/bindings.json
+++ b/Bindings/bindings.json
@@ -113,6 +113,24 @@
                     "label": "$eventHubTrigger_connection_label",
                     "help": "$eventHubTrigger_connection_help",
                     "placeholder": "[variables('selectConnection')]"
+                },
+                {
+                    "name": "cardinality",
+                    "value": "enum",
+                    "required": false,
+                    "defaultValue": "many",
+                    "enum": [
+                        {
+                            "value": "one",
+                            "display": "One"
+                        },
+                        {
+                            "value": "many",
+                            "display": "Many"
+                        }
+                    ],
+                    "label": "$eventHubTrigger_cardinality_label",
+                    "help": "$eventHubTrigger_cardinality_help"
                 }
             ]
         },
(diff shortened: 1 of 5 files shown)


Bindings/bindings.json
```diff
@@ -113,6 +113,26 @@
                     "label": "$eventHubTrigger_connection_label",
                     "help": "$eventHubTrigger_connection_help",
                     "placeholder": "[variables('selectConnection')]"
+                },
+                {
+                    "name": "cardinality",
+                    "value": "enum",
+                    "resource": "EventHub",
```
soninaren: Remove  "resource": "EventHub". This opening blades for resource selection like storage, eventhub connection

Bindings/bindings.json
```diff
@@ -113,6 +113,26 @@
...
+                            "display": "one"
+                        },
+                        {
+                            "value": "many",
+                            "display": "many"
+                        }
+                    ],
+                    "label": "$eventHubTrigger_cardinality_label",
+                    "help": "$eventHubTrigger_cardinality_help",
+                    "placeholder": "$selectCardinality"
```
soninaren: Don't need placeholder. It is only for textbox. This would be a dropdown.

Bindings/bindings.json
```diff
@@ -113,6 +113,26 @@
...
+                        {
+                            "value": "one",
+                            "display": "one"
+                        },
+                        {
+                            "value": "many",
+                            "display": "many"
+                        }
+                    ],
+                    "label": "$eventHubTrigger_cardinality_label",
```
soninaren: Add corresponding values for $eventHubTrigger_cardinality_label and  $eventHubTrigger_cardinality_help in resource files.
Here is a [sample ](https://github.com/Azure/azure-webjobs-sdk-templates/blob/dev/Resources/Resources.resx#L129).

mamaso: https://github.com/Azure/azure-webjobs-sdk-templates/pull/430/files?w=1#diff-73189aaa4df7ca5d944ae99dd2d5949aR1149

Didn't show up in the diff, funky.

PR #431, opened 2017-04-14
Adding HttpTriggerWith Parameters template

soninaren:


Templates/HttpTriggerWithParameters-CSharp/run.csx
```diff
@@ -0,0 +1,7 @@
+using System.Net;
+
+public static async Task<HttpResponseMessage> Run(HttpRequestMessage req,string name, TraceWriter log)
```
christopheranderson: Space between parameters

Add a comment explaining where "name" is coming from (via parameters)

alexkarcher-msft:
This looks great!
Ship it!

PR #484, opened 2017-06-29
[DocDB] adding partition keys bindings.json

### LINKED ISSUE #454: Add new DocDB partitionKey property to bindings.json
We've added it to function.json, we just need to add it to bindings.json. We also need a resource string and text for the label.

brettsam:
Fixes #454


brettsam:
Hey @soninaren -- can you take a look at this?

soninaren:
Looks good to me. Does runtime already recognize partition key property?

PR #510, opened 2017-08-01
AAD permissions for new button, minor renaming

glennamanns:
Added new button to create AAD app with appropriate permissions for a given binding or template. Corresponding UX PR: https://github.com/Azure/azure-functions-ux/pull/1615


Functions.Templates/Bindings/bindings.json
```diff
@@ -1755,6 +1755,17 @@
             "enabledInTryMode": false,
             "documentation": "$content=Documentation\\Token.md",
             "filters": ["MSGraph"],
+            "AADPermissions": [
```
soninaren: I am not sure if the permissions need to be configurable, but if they do need to be we would also need representative text for the scope being defined.

Functions.Templates/Templates/TimerTrigger-ExcelOutlook/metadata.json
```diff
@@ -0,0 +1,61 @@
+﻿{
+  "defaultFunctionName": "TimerTrigger-ExcelOutlook",
+  "description": "$Read two Excel tables, send emails based on the data, and then update an Excel table.",
+  "name": "Send Emails",
+  "language": "C#",
+  "enabledInTryMode": true,
+  "category": [ "$temp_category_msgraph" ],
+  "filters": [ "MSGraph" ],
+  "AADPermissions": [
```
soninaren: use UserPrompt field instead of adding the whole setting here again

Functions.Templates/Templates/TimerTrigger-ExcelOutlook/metadata.json
```diff
@@ -0,0 +1,61 @@
+﻿{
+  "defaultFunctionName": "TimerTrigger-ExcelOutlook",
+  "description": "$Read two Excel tables, send emails based on the data, and then update an Excel table.",
+  "name": "Send Emails",
```
soninaren: The name is very ambiguous and non descriptive, we already have templates that send emails. It would be okay to have a longer name if it needs to be.

Functions.Templates/Bindings/bindings.json
glennamanns: I added it as a special field because it's more like metadata than a setting. Settings are inputs/outputs, whereas this is metadata necessary for the AAD button. 

The permissions do not need to be configurable, as they are relatively constant and would only change if the MS Graph team changed their SDK.

soninaren: For help text, you can have a "i" icon that shows a tool tip similar to what we have for other things.

soninaren: I agree that, the settings are specific to input/output that get added to function.json. However anything outside of settings array applies to all the bindings and can possibly apply to all bindings. The Express AAD registration is a resource that needs to gets added specifically for MS Graph. 

The problem here is, it does not really belong in settings or outside of settings. So either we have to create a special section like related resources or find place that is closer to its behavior. The way i see it is, AAD is a resource needed only for MS Graph binding. Just like a storage account or eventhub. Would only change if MS Graph team changes SDK. So a more ideal location for it would be settings. and it also allows you take advantage of existing constructs that apply to settings in portal.

PR #511, opened 2017-08-01
Adding config names for all the settings

soninaren:


lindydonna:
Looks good, except it should be **WebHook**

PR #534, opened 2017-08-15
Fixing the CosmosDBTrigger-CSharp template to remove extraneous binding

soninaren:


soninaren:
re-opening the PR against master

PR #561, opened 2017-09-15
modifying strings for Graph bindings

mattchenderson:


Functions.Templates/Bindings/bindings.json
```diff
@@ -2808,8 +2606,8 @@
         {
           "name": "resource",
           "value": "string",
-          "label": "$userid_label",
-          "help": "$userid_help"
+          "label": "$MSGraphWebhook_in_listen_label",
```
soninaren: The label of the setting would be listen. Just wanted to make sure it was on purpose

PR #573, opened 2017-09-26
Updating VS templates

soninaren:


soninaren:
@ealsur can you take a look this. I have updated cosmos DB template for VS.

ealsur:
@soninaren Looks good 😄

PR #587, opened 2017-10-21
Names and descriptions are generic, add categoryStyle

agruning:


Functions.Templates/Resources/Resources.resx
```diff
@@ -1747,4 +1675,40 @@ function app.</value>
   <data name="temp_category_durableFunctions" xml:space="preserve">
     <value>Durable Functions</value>
   </data>
+  <data name="BlobTrigger_name" xml:space="preserve">
```
soninaren: Can't use resource strings for names, it breaks CLI, you see $values in the template names when you type func new, I am following up on that and am planning to remove the localization in 4 application insights templates that use resource strings in names

Functions.Templates/Resources/Resources.resx
```diff
@@ -1311,8 +1242,8 @@
   <data name="idToken_help_graphwebhook" xml:space="preserve">
     <value>Expression or explicit id token used for authenticating to MS Graph </value>
   </data>
-  <data name="ScheduledMailCSharp_description" xml:space="preserve">
-    <value>(Preview) A C# function that will periodically send emails</value>
+  <data name="ScheduledMail_description" xml:space="preserve">
+    <value>(Preview) A function that will periodically send emails</value>
```
soninaren: remove "(Preview)"

Functions.Templates/Resources/Resources.resx
agruning: Many function templates in dev have resource strings for names. How will they be localized without the resx?

PR #590, opened 2017-11-02
create experience updates

agruning:
updating some names and descriptions


Functions.Templates/Templates/DurableFunctionsActivity-CSharp/metadata.json
```diff
@@ -1,5 +1,5 @@
 {
-  "defaultFunctionName": "Hello",
+  "defaultFunctionName": "DurableFunctionsActivity",
```
soninaren: This function name is in line with function code [here.](https://github.com/Azure/azure-webjobs-sdk-templates/blob/v2.x/Functions.Templates/Templates/DurableFunctionsOrchestrator-CSharp/DurableFunctionsOrchestratorCSharp.cs)
Either undo the change or update the code as well.
Check with @cgillum for updates

cgillum: Yeah, the way this works is that orchestrator functions can call activity functions by name. Because portal templates can only produce one function at a time, the idea was that the orchestrator function template would be designed so that it could immediately call the activity function template. The orchestrator function template is a kind of "Hello world" template, so it made sense that it calls an activity function named "Hello". That's why we set the default function name as "Hello" for the activity function template (otherwise the user would need to edit the template before they could try it out).

Sorry if this is confusing - it might make more sense if I showed it to your in person. For now, I think it's probably best to leave the default function name as-is unless there is a strong reason to change it.

agruning: Gotcha. I was just changing it to make the format similar to the other templates, but I'll change it back

PR #617, opened 2018-01-05
Update Microsoft.NET.Sdk.Functions version to 1.0.7

ahmelsayed:


soninaren:
Do we need to update project template for v2.x as well?

Issue #599, opened 2017-11-10
Change display names for VS Tools


jeffhollan:
A number of customers get stuck by some of the display names on creating a new function.  "Connection" for example could potentially be more clear by naming it "Connection Key Name", "Path" for Event Hub should be called "Event Hub Name", etc.

paulbatum:
@jeffhollan Are you going to come up with a full list of proposed display name changes?

jeffhollan:
Yes thanks for ping. was looking for this item the other day. Will open a PR this week or next

soninaren:
This is done

PR #618, opened 2018-01-11
Renamed VS display names for properties

jeffhollan:
Not sure where the v2 templates are to make similar changes

### DIFF
--- a/Functions.Templates/Templates/BlobTrigger-CSharp/build.config/template.json
+++ b/Functions.Templates/Templates/BlobTrigger-CSharp/build.config/template.json
@@ -18,16 +18,16 @@
             "replaces": "Company.Function",
             "type": "parameter"
         },
-        "Connection": {
+        "ConnectionStringSetting": {
             "description": "The name of the app setting containing your Storage connection.",
             "type": "parameter",
-            "defaultValue": "",
+            "defaultValue": "AzureWebJobsStorage",
             "replaces": "ConnectionValue"
         },
         "Path": {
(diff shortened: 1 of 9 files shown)


jeffhollan:
Related to issue #599

paulbatum:
I'm not sure if this change actually achieves what you're looking for. There's a workflow you can use to test template changes, but I'm not sure where its documented (doesn't look like its in the repro wiki). Might need to sit on this change until @soninaren gets back.

soninaren:
@jeffhollan  the properties you are looking for vs-2017.3.host.json. The text property corresponds to the display text in visual studio. We would also need make the same changes in v2.x branch.
https://github.com/Azure/azure-webjobs-sdk-templates/blob/dev/Functions.Templates/Templates/BlobTrigger-CSharp/build.config/vs-2017.3.host.json#L23

All the documentation related to VS template should be in templating wiki for dotnet. I will some additional documentation in the repo wiki to make it easily accessible.

EhRom:
@DamiaoCastro, could this pull request can contain the fix for this issue https://github.com/Azure/Azure-Functions/issues/652#issuecomment-360980571 ?

DamiaoCastro:
@EhRom , this seems to be changes on the VS templates only. Not the functionality that generates the "function.json" file, if I'm not mistaken. Maybe that's part of another PR.

soninaren:
This changes in the PR has been already

soninaren:
@DamiaoCastro VS update should this https://github.com/Azure/Azure-Functions/issues/652 issue.

PR #635, opened 2018-02-22
eventGrid update

agruning:


Functions.Templates/Templates/EventGridTrigger-CSharp/metadata.json
```diff
@@ -6,7 +6,7 @@
     "category": [
         "$temp_category_experimental"
     ],
-    "categoryStyle": "eventHub",
+    "categoryStyle": "eventGrid",
```
soninaren: Eventhub and eventgrid have same icons now?

agruning: Yeah I will updating such that they each have their own

agruning: Yes, updating so they will be different

PR #641, opened 2018-03-03
Make default Python functions more pythonic

uiri:
The examples now automatically closes the opened file handlers using
Python's convenient `with` syntax


asavaritayal:
Is this related to Python support on v1?

uiri:
I saw some uses of `open(...).read()` which leaves a dangling, unclosed resource open. A `with open(...) as handle:` block ensures that the handle gets closed at the end of the block.

It looks like my branch is hopelessly out of date now, so I will close this PR.

PR #666, opened 2018-04-11
Relisting graph, fixing graph version

soninaren:


ConnorMcMahon:
Specifically, the Refresher should have the following code.
`#r "Microsoft.Azure.WebJobs.Extensions.Tokens"
#r "Microsoft.Azure.WebJobs.Extensions.O365"

using System.Net;
using Microsoft.AspNetCore.Mvc;
using Microsoft.Extensions.Primitives;
using Microsoft.Azure.WebJobs;

public static async Task Run(TimerInfo myTimer, UserSubscription[] existingSubscriptions, IBinder binder, TraceWriter log)
{
    log.Info($"C# Timer trigger function executed at: {DateTime.Now}");
    foreach (var subscription in existingSubscriptions)
    {
        // binding in code to allow dynamic identity
        var subscriptionsToRefresh = await binder.BindAsync<IAsyncCollector<string>>(
            new GraphWebhookSubscriptionAttribute()
            {
                Action = GraphWebhookSubscriptionAction.Refresh,
                Identity = TokenIdentityMode.UserFromId,
                UserId = subscription.UserId
            }
        );
        {
            log.Info($"Refreshing subscription {subscription.Id}");
            await subscriptionsToRefresh.AddAsync(subscription.Id);
        }
    }
}

public class UserSubscription {
    public string UserId {get; set;}
    public string Id {get; set;}
}
`

soninaren:
Have updated the code

PR #671, opened 2018-04-17
Fixing the OutlookMessageWebhookDeleter template

soninaren:


soninaren:
@ConnorMcMahon  can you review the fix?

PR #672, opened 2018-04-19
Adding iot hub template

jeffhollan:
First pass - couldn't figure out how to specify v1 vs v2, as v2 requires `Microsoft.Azure.WebJobs.Extensions.EventHubs` and not `Microsoft.Azure.WebJobs.ServiceBus`


soninaren:
Have added IotHub template in here https://github.com/Azure/azure-functions-templates/pull/674

PR #695, opened 2018-06-06
change package name according to the newest java function dependency

### LINKED ISSUE #694: Update the package name for Java function templates
The artifact of `azure-functions-java-core` has been renamed to `azure-functions-java-library`. And the package name has also been changed. We need to update the related templates as well.

jdneo:
resolve #694


jdneo:
@pragnagopa

PR #703, opened 2018-06-28
Add EventHub and EventGrid templates for Python

elprans:


Functions.Templates/Templates/EventGridTrigger-Python/function.json
```diff
@@ -0,0 +1,10 @@
+{
+  "bindings": [
+    {
+      "type": "eventGridTrigger",
+      "name": "event",
+      "direction": "in"
+    }
+  ],
+  "disabled": false
```
asavaritayal: Remove "disabled":false as the V2 runtime does not respect it anymore.

elprans: It is present in all other templates though.  Should it be removed everywhere for consistency?

asavaritayal: Yes, it needs to be removed everywhere else as well. I'll create a PR for that separately.

elprans:
@asavaritayal Fixes pushed.

asavaritayal:
@soninaren can we go ahead and merge this?

PR #707, opened 2018-07-10
Swapping out TraceWriter with ILogger

jeffhollan:
Adding in reference to this issue: https://github.com/Azure/azure-functions-templates/issues/667


jeffhollan:
https://github.com/Azure/azure-functions-templates/issues/667

PR #708, opened 2018-07-24
Fixing the malformed namespace issue

soninaren:


soninaren:
Fixes https://github.com/Azure/azure-functions-templates/issues/487

Functions.Templates/Templates/BlobTrigger-CSharp/build.config/template.json
```diff
@@ -16,7 +16,15 @@
         "namespace": {
             "description": "namespace for the generated code",
             "replaces": "Company.Function",
-            "type": "parameter"
+            "type": "parameter",
+            "forms": {
```
mlorbetske: Per the mail, none of these changes to the item templates should be made, adding the `safe_namespace` form is not required or desired for this use. Only the change to the project files are required

soninaren: Have fixed the PR

PR #710, opened 2018-07-24
Delete Java templates 

pragnagopa:
These are copies from:
https://github.com/Microsoft/azure-maven-plugins/tree/develop/azure-functions-maven-plugin and are not used anywhere


pragnagopa:
@jdneo , @EricJizbaMSFT - Please ACK that these templates are not used in VSCode
@ahmelsayed - Does Core tools have dependency on these?

jdneo:
It's not used in VS Code extension. Please go ahead to remove them.

PR #718, opened 2018-08-08
[JavaScript] making templates async

jeffhollan:
[Per this issue](https://github.com/Azure/azure-functions-nodejs-worker/issues/94#issuecomment-411240990)

### DIFF
--- a/Functions.Templates/Templates/BlobTrigger-JavaScript/index.js
+++ b/Functions.Templates/Templates/BlobTrigger-JavaScript/index.js
@@ -1,4 +1,3 @@
-module.exports = function (context, myBlob) {
+module.exports = async function (context, myBlob) {
     context.log("JavaScript blob trigger function processed blob \n Name:", context.bindingData.name, "\n Blob Size:", myBlob.length, "Bytes");
-    context.done();
 };
--- a/Functions.Templates/Templates/CosmosDBTrigger-JavaScript/index.js
+++ b/Functions.Templates/Templates/CosmosDBTrigger-JavaScript/index.js
@@ -1,6 +1,5 @@
-module.exports = function (context, documents) {
+module.exports = async function (context, documents) {
     if (!!documents && documents.length > 0) {
         context.log('Document Id: ', documents[0].id);
     }
-    context.done();
 }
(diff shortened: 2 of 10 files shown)


mhoeger:
My only concern is that people will arbitrarily try to make their code not async and run into timeout errors when context.done() is missing. 

I think this is a more edge-case scenario though, and exporting async functions is overwhelmingly a better pattern. Also long-term shouldn't be too much of an issue when we resolve this :) https://github.com/Azure/azure-functions-nodejs-worker/issues/106

jeffhollan:
@mhoeger I'm still learning about this `Promise` stuff with node so clarifying I understand.  Would risk be something that in my function I write some code like:

`axios.post(process.env.myUrl)` instead of `await axios.post(process.env.myUrl)` which if the function is `async` would not execute as expected, whereas if not async and had `context.done()` the former would work as expected?

mhoeger:
@jeffhollan - the risk i was talking about is that removing `context.done()` from the template decreases the chances that a new user will know about context.done() as a thing that is sometimes needed. Just pointing out that we may see more users report or experience "function timeout errors with no cause" due to lack of exposure to the context.done callback, but think it's ok and solvable by searching docs.

Using `axios.post(process.env.myUrl)` without await in an `async` function will execute the same as `axios.post(process.env.myUrl); context.done();` so no worries there :)

jeffhollan:
Ah ok so I think if I understand right the potential risk is someone looks at template and says "I don't need this to be async" and so removes that word, but without knowing they now need to add `context.done()` may be confused as to why their function execution isn't completing.  That's a fair point.

soninaren:
There would be a middle ground though, we can leave some templates as is. May be http can be as is. So when users run into timeout issues they would find out why one function is working and other isn't

paulbatum:
I feel like a simple doc page that has an example of a js function that use context.done and another that is async and a little bit of explanation would address the confusion and let us keep async templates with very little downside.

jeffhollan:
@ColbyTresness - I believe you're on the hook to clean up and update the JavaScript docs before v2 GA.  Can you do as @paulbatum outlined above?  If we see support tickets pop up we could consider adding some `// comments` to the template as well. I'd prefer to keep them consistent though

ColbyTresness:
Yes @jeffhollan @paulbatum , this is definitely something we plan to address in our docs improvements for V2. @kashimiz @mhoeger and I have done a pretty comprehensive read through of the JS docs and have identified a bunch of areas for improvement, including the confusion around context.done.

Issue #690, opened 2018-05-31
Log shows executed successfully for Face locator template although computer vision key is invaild


SimonGuo009:
**Portal Environment:**  https://ms.portal.azure.com
**Visual Studio:**  2017 Enterprise 15.7.2
**Azure Function Tools:**  15.0.40502.0
**Browser:**  English/Chrome

Please refer to snapshot:
![faceinvalidkey](https://user-images.githubusercontent.com/18321627/40777023-13a573c8-64bc-11e8-8621-3f9ba3168e45.png)

**Repro steps:**
1. Create a V1 Face locator template in VS.
2. Type an invalid computer vision key.
3. Run it locally then trigger it, verify the logs.

paulbatum:
Hmm the template should also use a static http client

PR #719, opened 2018-08-08
fixing face API issue #690

jeffhollan:


jeffhollan:
per [this issue](https://github.com/Azure/azure-functions-templates/issues/690)

PR #723, opened 2018-08-20
fixing Event Hub reference

jeffhollan:
per breaking changes rolling out


paulbatum:
This also needs a bump to the extension version line from `"version": "3.0.0-beta5"` to beta7 in metadata.json

jeffhollan:
Thanks - at first I assumed @soninaren had a separate branch with the version numbers but after looking can see that's not the case. Pushed a commit to also update NuGet version.

soninaren:
References have already been updated in this PR, https://github.com/Azure/azure-functions-templates/pull/717

jeffhollan:
But expect that version won't be pushed into `dev` as it has the runtime version appendage for MyGet correct?  Either way think we have right work being tracked between 2 PRs.  @soninaren let me know if you want me to make any changes or use your other branch for these changes to make your life easier.

paulbatum:
@soninaren we can close this right? Its covered by your other PR...

PR #738, opened 2018-09-07
Major F# templates for compiled libraries

cartermp:
There are some templates that are script-based, and won't run on v2 because FSI doesn't work properly on .NET Core. However, compiled library-style templates work just fine.

This adds some of them (major templates, at least).


forki:
Ping

paulbatum:
I'm looking at the output from this build:
https://ci.appveyor.com/project/appsvc/azure-webjobs-sdk-templates/builds/23480548/artifacts

To validate, I downloaded the build artifacts, then ran:

dotnet new --install "D:\localnuget\Microsoft.Azure.WebJobs.ItemTemplates.2.0.10348-ikqanjqc.nupkg"

I confirmed this installed the C# item templates, but the F# item templates don't appear. I cracked open the NuGet and confirmed that the F# templates are not present.

I suspect the PR needs to update this file:
https://github.com/Azure/azure-functions-templates/blob/dev/Functions.Templates/ItemTemplates.nuspec

I am not exactly the expert in this area so let me know if there is an important detail I'm missing..

cartermp:
As of latest commit I now get them to show up in ItemTemplates and they appear under `dotnet new`:
image

However I cannot create any projects in either C# or F# even if I install the project templates nupkg. Any suggestions?
image

paulbatum:
Okay, after installing the output from the latest build, I'm able to do this:

`dotnet new func -lang F# -n fsharptest`

And then:
`dotnet new HttpTrigger -lang F# -n httptrigger1`

This does NOT add `httptrigger1.fs` to the project file. So I manually added this:

```xml
  <ItemGroup>
    <Compile Include="httptrigger1.fs" />
  </ItemGroup>
```

Then I run: `dotnet build`

I then switch to the build output dir at `bin\Debug\netcoreapp2.1` and confirmed that it includes my `httptrigger1\function.json` file. 

I run `func host start` at that location and it picks up my function. If I try to run it I get an error, because I think there is a bug with the http template:

```
F# HTTP trigger function processed a request.
[4/3/2019 6:15:07 PM] Executed 'httptrigger1' (Failed, Id=2068014a-6331-41f9-ad0c-b52d48dd1acc)
[4/3/2019 6:15:07 PM] System.Private.CoreLib: Exception while executing function: httptrigger1. Microsoft.Extensions.Primitives: Index was outside the bounds of the array.
fail: Microsoft.AspNetCore.Server.Kestrel[13]
      Connection id "0HLLOC16DQOBQ", Request id "0HLLOC16DQOBQ:00000001": An unhandled exception was thrown by the application.
Microsoft.Azure.WebJobs.Host.FunctionInvocationException: Exception while executing function: httptrigger1 ---> System.IndexOutOfRangeException: Index was outside the bounds of the array.
   at Microsoft.Extensions.Primitives.StringValues.get_Item(Int32 index)
   at Company.Function.httptrigger1.run(HttpRequest req, ILogger log) in D:\code\temp\fsharp-templates\httptrigger1.fs:line 17
```

My F# is bad, but it looks like this is the problem:

```fs
let name = req.Query.["name"].[0]
```

So, outstanding issues:
1. Is there some way that the item templates can automatically be added to the .fsproj?
2. The http template should not blow up if you pass no params.

Angr1st:
@cartermp  I sent a fix for the HTTPTrigger

cartermp:
@paulbatum Not sure about this:

> Is there some way that the item templates can automatically be added to the .fsproj?

Since F# has file ordering requirements, we cannot use globbing like C#. So I think there needs to be some kind of task that adds it to an ItemGroup (and for subsequent files, adds to that same ItemGroup)

TclasenITVT:
@cartermp does your template work with query parameters?

cartermp:
@TclasenITVT Yes (though I want to flesh out the other templates as well, I'm sure they could fail pretty easily)

paulbatum:
I don't think we need to treat the project file problem as a blocker.

@soninaren does this look OK to go in? (Don't worry about reviewing the individual templates)

soninaren:
Merged the commits after a few fixes and resolving merge conflicts.
Here is the commit id.
https://github.com/Azure/azure-functions-templates/commit/32ad34d09ea3208659b8db0b1568e5c2212cceaa

paulbatum:
Thank you @cartermp and @Angr1st for your help with this, sorry it took us a while to get this in!

dsyme:
Great to see!

Are there more follow up items we need to do for F# templates for Azure Functions?

paulbatum:
See some of the conversation above - the current experience requires the user to manually update their .fsproj after adding a new function. Is that pretty normal expectation for a F# user?

cartermp:
@paulbatum Yes and no. For a new project, a file is added to the project by default (such as `Library.fs`) for a library proejct.

Subsequent adding of project files is typically done with IDE tools. Examples:

image

In the VS project system, we implement certain ordering semantics such that ading a file to a given scope (project, folder, subfolder, etc.) does the right thing by keeping it in that scope. If Add Above/Below is not chosen the file is added to the top of the set within that scope.

In VSCode, there is a less intricate system via the Ionide plugin:

image

But a lot of F# users with VSCode manually add to the project file.

For the purposes of templates, I'd think that the F# file for a given template should be added to that project file.

PR #751, opened 2018-09-25
Update Event Grid Categories

### LINKED ISSUE #747: JS Event Grid create marked as experimental
image

agruning:
fixes #747

### DIFF
--- a/Functions.Templates/Templates/EventGridTrigger-Python/metadata.json
+++ b/Functions.Templates/Templates/EventGridTrigger-Python/metadata.json
@@ -4,7 +4,8 @@
     "name": "Event Grid trigger",
     "language": "Python",
     "category": [
-        "$temp_category_experimental"
+        "$temp_category_core",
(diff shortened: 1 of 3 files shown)


Functions.Templates/Templates/EventGridTrigger-Python/metadata.json
```diff
@@ -4,7 +4,8 @@
     "name": "Event Grid trigger",
     "language": "Python",
     "category": [
-        "$temp_category_experimental"
```
mathewc: Python is an experimental language though (not GA), so shouldn't this remain experimental?

agruning: Actually none of the python templates have the category. We shouldn't be using category "experimental" at all in V2.

mathewc: Ah, ok.

PR #759, opened 2018-10-01
Updating HttpTrigger template formatting

fabiocav:


fabiocav:
@soninaren I'm seeing a build failures, but I don't think it's related to my changes. Can you confirm?

PR #761, opened 2018-10-02
Table binding via Portal + docs for multi-row filtering

maiqbal11:
Adding table binding support for Portal. Also added documentation and an example for doing multi-row filtering using JArray binding since we are not longer support IQueryable.

### DIFF
--- a/Functions.Templates/Documentation/tableIn.md
+++ b/Functions.Templates/Documentation/tableIn.md
@@ -13,6 +13,10 @@
 
 The `table` binding can serialize or deserialize objects in JavaScript or C# functions. The objects will have RowKey and PartitionKey properties. 
 
+In C# functions, you can also bind to the following type:
+
+ * `T` where `T` implements `ITableEntity`
+
 #### Storage tables binding scenarios
 
 The table binding supports the following scenarios:
--- a/Functions.Templates/Documentation/tableIn.md
+++ b/Functions.Templates/Documentation/tableIn.md
@@ -13,11 +13,6 @@
 
 The `table` binding can serialize or deserialize objects in JavaScript or C# functions. The objects will have RowKey and PartitionKey properties. 
 
-In C# functions, you can also bind to the following types:
-
-* `T` where `T` implements `ITableEntity`
-* `IQueryable<T>` 
-
 #### Storage tables binding scenarios
 
 The table binding supports the following scenarios:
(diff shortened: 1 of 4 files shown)


Functions.Templates/Documentation/tableIn.md
```diff
@@ -13,10 +13,9 @@
 
 The `table` binding can serialize or deserialize objects in JavaScript or C# functions. The objects will have RowKey and PartitionKey properties. 
 
-In C# functions, you can also bind to the following types:
+In C# functions, you can also bind to the following type:
 
-* `T` where `T` implements `ITableEntity`
-* `IQueryable<T>` 
+ * `T` where `T` implements `ITableEntity`
```
mathewc: There is also a CloudTable binding - you might as well mention that here as well

Functions.Templates/Documentation/tableIn.md
```diff
@@ -64,25 +59,32 @@ module.exports = function (context, myQueueItem) {
 };
 ```
 
-#### Storage tables example: C# example that reads multiple table entities
+#### Storage tables example: C# example that reads multiple table entities based on a queue trigger
```
mathewc: You didn't add the "based on queue trigger" above, just here. Why?

Functions.Templates/Documentation/tableIn.md
```diff
@@ -64,25 +59,32 @@ module.exports = function (context, myQueueItem) {
 };
 ```
 
-#### Storage tables example: C# example that reads multiple table entities
+#### Storage tables example: C# example that reads multiple table entities based on a queue trigger
 
-The C# code adds a reference to the Azure Storage SDK so that the entity type can derive from `TableEntity`.
+The C# code adds a reference to Newtonsoft JSON library so that we can utilize JSON constructs to process multiple table entries. The `queuePerson` argument is bound to a queue entry while the `inputTable` is bound to an Azure Storage Table. We can specify the `filter` property so that, for example, only table results for persons older than the person in the queue are returned ( specify `"Age gt {Age}"` where `{Age}` represents the age of the person from the queue).
```
mathewc: nit: extra space in "( specify"

Functions.Templates/Documentation/tableIn.md
```diff
@@ -64,25 +59,32 @@ module.exports = function (context, myQueueItem) {
...
 
-public static void Run(string myQueueItem, IQueryable<Person> tableBinding, ILogger log)
+public static void Run(Person queuePerson, Newtonsoft.Json.Linq.JArray inputTable, ILogger log)
 {
-    log.LogInformation($"C# Queue trigger function processed: {myQueueItem}");
-    foreach (Person person in tableBinding.Where(p => p.PartitionKey == myQueueItem).ToList())
+    foreach (Newtonsoft.Json.Linq.JToken jToken in inputTable)
     {
-        log.LogInformation($"Name: {person.Name}");
+        Person p = Newtonsoft.Json.JsonConvert.DeserializeObject<Person>(jToken.ToString());
```
mathewc: You should be able to bind to a Person[] directly rather than JArray. Doesn't that work?

maiqbal11: Just tried this using both the WebJobs project and the Portal code editor. The WebJobs project is showing this an invalid binding while the Portal code is failing silently even though it compiles (I imagine that this is because of some background failure similar to the one I saw in the WebJobs project).

Functions.Templates/Documentation/tableIn.md
mathewc: Yeah, I'd just remove it

PR #764, opened 2018-10-04
added default proxies.json

joescars:
Added empty proxies.json file to C# & F# templates with 'Copy to Output Directory' set to 'Copy if Newer' . IMO more developer friendly as the file is already there and being published. This way, when they add proxies after the functions have already been built it is already configured. Additionally they can just delete the file if they are not using.


soninaren:
@jeffhollan , @paulbatum any thoughts?

alexkarcher-msft:
This looks super awesome to me! As long as we've verified that it won't break the portal I'm all good to merge this in.
It *should* have no effect on the portal, and will make it easier for VS and CLI users to get started with proxies 😃

joescars:
@alexkarcher-msft to clarify this PR does not add the proxies to the GUI as noted in #766 . It just creates a default proxies.json that publishes on build when you create a new Azure Functions project.

soninaren:
Closing the PR based on the comment from @alexkarcher-msft , Will follow up on issue #766.

PR #765, opened 2018-10-05
Adding azure branding to the template title

soninaren:


soninaren:
@ggailey777 can you review it as well?

PR #767, opened 2018-10-05
Removing languages from all default template names

soninaren:


soninaren:
https://github.com/Azure/azure-functions-templates/issues/762

soninaren:
@ggailey777 can you review this one as well?

PR #770, opened 2018-10-10
Add Easy Auth local settings file to git ignore, project template

glennamanns:
This is part of a larger goal to enable Easy Auth to run locally. The relevant Azure Functions CLI PR is here: https://github.com/Azure/azure-functions-core-tools/pull/789 

The CLI PR is currently blocked/waiting on the Easy Auth Linux+Windows combine, and this templates PR should be merged about the same time as that one.


soninaren:
@glennamanns , I don't think this is the right change. We shouldn't be adding an additional files to the default template which is an opt in feature. Instead this should be a CLI command like proxies. `Function proxy create --name <yourproxyname>`.

glennamanns:
Closing PR based on @soninaren's comment. I will instead programmatically add local.middleware.json to the Function's .csproj during the create-aad command.

PR #845, opened 2019-03-07
Wrong casing in TimerTrigger JavaScript/Typescript `isPastDue`

mhoeger:
isPastDue should be IsPastDue. In V1, there used to be hard coded logic to make this lower case (https://github.com/Azure/azure-functions-host/blob/2a73a53260b55e8abc61c34da39be04244e5a957/src/WebJobs.Script/Description/Node/NodeFunctionInvoker.cs#L406). In V2, this custom logic doesn't happen.


anthonychu:
Looks good.

If it makes sense to change this at the same time, could you please also fix the inconsistent formatting? (Space between `if` and `(` and move the open brace to the same line) Thanks!

PR #853, opened 2019-03-15
Add SignalR Service negotiate endpoint template

anthonychu:
Add templates for creating a SignalR Service negotiation HTTP Trigger

This PR is for portal only. Will need some help to add VS in another PR. @soninaren 

https://github.com/Azure/azure-functions-templates/issues/843


mhoeger:
The TypeScript looks good to me! @soninaren - if you're doing the deployment, can you keep me in the loop? want to verify portal + typescript changes

anthonychu:
@soninaren Checking in to see if we can get this merged. Thanks.

PR #856, opened 2019-03-27
Project template name looks better with a space

maartenba:
Currently looks quite ugly when running `dotnet new-l`:

```
Templates                                         Short Name           Language          Tags
------------------------------------------------------------------------------------------------------------------------------
AzureFunctions                                    func                 [C#], F#          Azure Functions/Solution
Console Application                               console              [C#], F#, VB      Common/Console
Class library                                     classlib             [C#], F#, VB      Common/Library
...
```

Same in graphical tools:

image


soninaren:
Thanks @maartenba for the PR.

PR #865, opened 2019-05-19
Provide default placeholder values for use with dotnet new

maartenba:
Makes the template easier to use with `dotnet new`


maartenba:
Any updates on this one?

PR #879, opened 2019-06-10
EN strings.json seeding for new template localization mechanism for Visual Studio.

phenning:


phenning:
Initial en versions of the strings.json files which are required for the localization mechanism in VS 16.3. Localized versions of these files should be shipped with each template in a ja-JP, etc sub folder under the vs-2017.3.host.json file location.

fabiocav:
@phenning this is a pretty old PR and I'm wondering if it is still applicable. Did you sync with @soninaren offline? Thanks!

phenning:
There is actually a newer localization mechanism supported in Dev17, we can abandon this and work on getting that implemented instead in the future/

PR #881, opened 2019-06-13
#870 python http template

t-eckert:
This PR addresses issue #870 HTTP - Template Improvements for Python. 

- Code in `__init__.py` is more pythonic.
- Adds README with resources.
- Uses [Black](https://github.com/python/black), standard Python formatting.

Please let me know if this is what you are looking for as I continue work on other templates. I want to make your jobs easier and help the transition out of preview to go as smoothly as possible.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,5 @@ Publish
 tools
 /node_modules
 /.output
+
+.env
--- a/Functions.Templates/Templates/HttpTrigger-Python/README.md
+++ b/Functions.Templates/Templates/HttpTrigger-Python/README.md
@@ -1,0 +1,54 @@
+# Python HttpTrigger
+
+The Azure Functions Python HTTP Trigger creates an endpoint: `<function-host>/api/<function-name>`. When an HTTP request is made against that endpoint, the `main()` function in `__init__.py` is called.
+
+> Throughout this documentation variables in angle brackets `<>` are placeholders and will be different in your application.
+
+## Default Structure
+
+``` text
+<function-name>/      # directory holding all files for the HttpTrigger Function
+└── __init__.py     # this module will be called when the Function is run
+    function.json   # defines that __init__.py will be called and which bindings are used
+    sample.dat      # contains a sample input for the Function
+```
+
+## Testing the Template Function
+
+The default function template when a new HttpTrigger is created takes a value for `name` and returns `Hello <name>!`. To test this functionality, from the root of the Azure Function App, execute
+
+``` bash
+func host start
+```
+
+Once running, you should see
+
+``` text
+Http Functions:
+
+        <function-name>: [GET,POST] http://localhost:7071/api/<function-name>
+```
+
+Open the URL in a browser. You should get the response
+
+``` text
+Please pass a name on the query string or in the request body
+```
+
+To the URL, append `?name=Sophia` so it reads
+
+``` text
+http://localhost:7071/api/<function-name>?name=Sophia
+```
+
+You should now get the response
+
+``` text
+Hello Sophia!
+```
+
+## Useful Resources
+
+- Software: [Postman](https://www.getpostman.com/) is a cross-platform application for testing HTTP requests.
+- Guide: [Create an HTTP triggered function in Azure](https://docs.microsoft.com/en-us/azure/azure-functions/functions-create-first-function-python)  
+- Examples: [Serverless Library - Python](https://serverlesslibrary.net/?technology=Functions%202.x&language=Python)
--- a/Functions.Templates/Templates/HttpTrigger-Python/README.md
+++ b/Functions.Templates/Templates/HttpTrigger-Python/README.md
@@ -1,21 +1,23 @@
 # Python HttpTrigger
 
-The Azure Functions Python HTTP Trigger creates an endpoint: `<function-host>/api/<function-name>`. When an HTTP request is made against that endpoint, the `main()` function in `__init__.py` is called.
+The HTTP trigger lets you invoke a function with an HTTP request. You can use an HTTP trigger to build serverless APIs and respond to webhooks. It creates an endpoint like  `https://<app-name>.azurewebsites.net/api/api/<function-name>`. When an HTTP request is made against that endpoint, the `main()` function in `__init__.py` is called.
+
 
 > Throughout this documentation variables in angle brackets `<>` are placeholders and will be different in your application.
 
 ## Default Structure
 
 ``` text
-<function-name>/      # directory holding all files for the HttpTrigger Function
+<function-name>/    # directory holding all files for the HttpTrigger Function
 └── __init__.py     # this module will be called when the Function is run
     function.json   # defines that __init__.py will be called and which bindings are used
+    README.md       # documentation for using HTTP Trigger
     sample.dat      # contains a sample input for the Function
 ```
 
-## Testing the Template Function
+## Run the Function Locally
 
-The default function template when a new HttpTrigger is created takes a value for `name` and returns `Hello <name>!`. To test this functionality, from the root of the Azure Function App, execute
+The default function template when a new HttpTrigger is created takes a value for `name` and returns `Hello <name>!`. If you're running in VS Code, you can press F5 to test your function locally. Otherwise, from the root of the Azure Function App, execute
 
 ``` bash
 func host start
@@ -47,8 +49,8 @@ You should now get the response
 Hello Sophia!
 ```
 
-## Useful Resources
+## Resources
 
-- Software: [Postman](https://www.getpostman.com/) is a cross-platform application for testing HTTP requests.
-- Guide: [Create an HTTP triggered function in Azure](https://docs.microsoft.com/en-us/azure/azure-functions/functions-create-first-function-python)  
-- Examples: [Serverless Library - Python](https://serverlesslibrary.net/?technology=Functions%202.x&language=Python)
+- [Azure Functions Python developer reference](https://aka.ms/python-functions-guide)  
+- [HTTP triggers and bindings](https://docs.microsoft.com/azure/azure-functions/functions-bindings-http-webhook)
+- [Serverless Library](https://aka.ms/python-functions-samples)
--- a/Functions.Templates/Templates/HttpTrigger-Python/__init__.py
+++ b/Functions.Templates/Templates/HttpTrigger-Python/__init__.py
@@ -4,21 +4,21 @@ import azure.functions as func
 
 
 def main(req: func.HttpRequest) -> func.HttpResponse:
-    logging.info('Python HTTP trigger function processed a request.')
+    logging.info("Python HTTP trigger function processed a request.")
 
-    name = req.params.get('name')
+    name = req.params.get("name")
     if not name:
         try:
             req_body = req.get_json()
         except ValueError:
             pass
         else:
-            name = req_body.get('name')
+            name = req_body.get("name")
 
     if name:
         return func.HttpResponse(f"Hello {name}!")
     else:
         return func.HttpResponse(
-             "Please pass a name on the query string or in the request body",
-             status_code=400
+            "Please pass a name on the query string or in the request body",
+            status_code=400,
         )
--- a/Functions.Templates/Templates/HttpTrigger-Python/__init__.py
+++ b/Functions.Templates/Templates/HttpTrigger-Python/__init__.py
@@ -8,10 +8,10 @@ def main(req: func.HttpRequest) -> func.HttpResponse:
 
     name = None
     try:
-        # Get value for `name` from parameters or JSON body
+        # Get value for `name` from parameters or JSON body.
         name = req.params.get("name") or req.get_json().get("name")
     except ValueError:
-        pass
+        name = None
 
     return (
         func.HttpResponse(f"Hello {name}!")
--- a/Functions.Templates/Templates/HttpTrigger-Python/__init__.py
+++ b/Functions.Templates/Templates/HttpTrigger-Python/__init__.py
@@ -6,14 +6,11 @@ import azure.functions as func
 def main(req: func.HttpRequest) -> func.HttpResponse:
     logging.info("Python HTTP trigger function processed a request.")
 
-    name = req.params.get("name") or req.get_json().get("name")
-    if not name:
-        try:
-            req_body = req.get_json()
-        except ValueError:
-            pass
-        else:
-            name = req_body.get("name")
+    try:
+        # Get value for `name` from parameters or JSON body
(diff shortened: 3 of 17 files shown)


Functions.Templates/Templates/HttpTrigger-Python/__init__.py
```diff
@@ -4,21 +4,19 @@
...
-            req_body = req.get_json()
-        except ValueError:
-            pass
-        else:
-            name = req_body.get('name')
+    try:
+        # Get value for `name` from parameters or JSON body
+        name = req.params.get("name") or req.get_json().get("name")
+    except ValueError:
+        pass
```
zooba: `name` will be undefined in this path, whereas before it would have been `None`

t-eckert: Should I instead say

``` python
name = req.params.get("name") or req.get_json().get("name") or None
```

t-eckert: Or, nevermind as the error thrown will mean that is never set. I'll keep at it.

t-eckert: The easy way out is to just write 

``` python 
name = None
try:
	name = req.params.get("name") or req.get_json().get("name")
except ValueError:
    pass
```

t-eckert: Just seems sloppy to me.

zooba:
```suggestion
        name = None
```

This will ensure you've assigned something to `name`. Alternatively, you could put `name = None` _before_ the `try` statement, but this is probably neater since you need the `pass` here anyway.

t-eckert: I like this change.

Functions.Templates/Templates/HttpTrigger-Python/README.md
```diff
@@ -0,0 +1,54 @@
+# Python HttpTrigger
+
+The Azure Functions Python HTTP Trigger creates an endpoint: `<function-host>/api/<function-name>`. When an HTTP request is made against that endpoint, the `main()` function in `__init__.py` is called.
+
+> Throughout this documentation variables in angle brackets `<>` are placeholders and will be different in your application.
+
+## Default Structure
```
t-eckert: Reminder to self- rename to Template Structure

.gitignore
```diff
@@ -16,4 +16,6 @@ Publish
 /packages
 tools
 /node_modules
-/.output
\ No newline at end of file
+/.output
+
+.env
```
brettcannon: Is this to avoid a virtual environment or [environment definition files](https://code.visualstudio.com/docs/python/environments#_environment-variable-definitions-file)?

Functions.Templates/Templates/HttpTrigger-Python/__init__.py
```diff
@@ -4,21 +4,19 @@
...
-    name = req.params.get('name')
-    if not name:
-        try:
-            req_body = req.get_json()
-        except ValueError:
-            pass
-        else:
-            name = req_body.get('name')
+    try:
+        # Get value for `name` from parameters or JSON body
```
brettcannon:
```suggestion
        # Get value for `name` from parameters or JSON body.
```

Functions.Templates/Templates/HttpTrigger-Python/__init__.py
```diff
@@ -4,21 +4,19 @@
...
+        name = req.params.get("name") or req.get_json().get("name")
+    except ValueError:
```
brettcannon: What is throwing `ValueError`? If `req.params` is a dict then that won't do it. Is it because `req.get_json()` might not exist? If the latter is true then you could also do:
```python
name = req.params.get("name") or getattr(req, "get_json", lambda: {})().get("name")
```
Not sure if that's really readable, but otherwise the `try` block seems overly broad.

Functions.Templates/Templates/HttpTrigger-Python/__init__.py
```diff
@@ -4,21 +4,19 @@
...
+        name = None
 
-    if name:
-        return func.HttpResponse(f"Hello {name}!")
-    else:
-        return func.HttpResponse(
-             "Please pass a name on the query string or in the request body",
-             status_code=400
+    return (
```
brettcannon: I personally prefer the more explicit approach that was there rather than the ternary operator since it's so long and you don't realize what's going on until you read all the lines.

.gitignore
t-eckert: I created a virtual environment to keep development packages (Black, etc.) but didn't want to commit it to git.

Functions.Templates/Templates/HttpTrigger-Python/__init__.py
t-eckert: It's because `get_json()` might not exist. I agree about the `try` block. I was hoping to avoid it if possible.

Functions.Templates/Templates/HttpTrigger-Python/__init__.py
t-eckert: I trust you on readability and "Pythonicness". I'll revert to the explicit.

.gitignore
brettcannon: So the common/recommended practice is to name the environment `.venv` so as to not conflict with environment variable definition files. (I'm also noticing while I read the Functions docs that they unfortunately document `.env`.)

t-eckert: Oh! I learned it from them.

Functions.Templates/Templates/HttpTrigger-Python/__init__.py
t-eckert: You can see the error being raised here: https://github.com/Azure/azure-functions-python-worker/blob/dev/azure/functions_worker/bindings/http.py

t-eckert: Lines 49 through 59

brettcannon:
```python
name = req.params.get("name")  # If this was Python 3.8 you could use assignment expressions 😁 .
if not name:
    try:
        json_data = req.get_json()
    except AttributeError:  # You had ValueError, but you said "might not exist".
        pass
    else:
        name = json_data.get("name")
```
Or:
```python
name = req.params.get("name")
if not name and hasattr(req, "get_json"):  # Once again, assignment expression would be nice. 😁 
    name = req.get_json().get("name")
```

t-eckert: I like the latter. I will implement that one.

.gitignore
brettcannon: [pipenv uses .venv](https://pipenv.readthedocs.io/en/latest/advanced/#custom-virtual-environment-location) if you set `PIPENV_VENV_IN_PROJECT`.

asavaritayal: Virtual environments are no longer required for Python Functions development. However, we still recommend using them. There's a WI on me to update the guidance in the docs. Given that there's still a mix of folks using both `.env' and `.venv', my suggestion would be to include both here if there's no downside.

Functions.Templates/Templates/HttpTrigger-Python/README.md
```diff
@@ -0,0 +1,54 @@
+# Python HttpTrigger
+
+The Azure Functions Python HTTP Trigger creates an endpoint: `<function-host>/api/<function-name>`. When an HTTP request is made against that endpoint, the `main()` function in `__init__.py` is called.
```
asavaritayal:
```suggestion
The HTTP trigger lets you invoke a function with an HTTP request. You can use an HTTP trigger to build serverless APIs and respond to webhooks. It creates an endpoint like  `https://<app-name>.azurewebsites.net/api/api/<function-name>`. When an HTTP request is made against that endpoint, the `main()` function in `__init__.py` is called.
```

Functions.Templates/Templates/HttpTrigger-Python/README.md
asavaritayal: I think "Default" is fine.

Functions.Templates/Templates/HttpTrigger-Python/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+> Throughout this documentation variables in angle brackets `<>` are placeholders and will be different in your application.
+
+## Default Structure
+
+``` text
+<function-name>/      # directory holding all files for the HttpTrigger Function
+└── __init__.py     # this module will be called when the Function is run
+    function.json   # defines that __init__.py will be called and which bindings are used
+    sample.dat      # contains a sample input for the Function
```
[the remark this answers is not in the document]
asavaritayal: You could also leave it out since it's not integral to the working of the function. Up to you.

Functions.Templates/Templates/HttpTrigger-Python/README.md
```diff
@@ -0,0 +1,54 @@
...
+```
+
+## Testing the Template Function
```
asavaritayal:
```suggestion
## Run the function locally
```

Functions.Templates/Templates/HttpTrigger-Python/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+The default function template when a new HttpTrigger is created takes a value for `name` and returns `Hello <name>!`. To test this functionality, from the root of the Azure Function App, execute
+
+``` bash
+func host start
+```
+
```
asavaritayal:
```suggestion
If you're running in VS Code, you can simply press F5 to test your function locally. 
```

Functions.Templates/Templates/HttpTrigger-Python/README.md
```diff
@@ -0,0 +1,54 @@
...
+You should now get the response
+
+``` text
+Hello Sophia!
+```
+
+## Useful Resources
+
+- Software: [Postman](https://www.getpostman.com/) is a cross-platform application for testing HTTP requests.
+- Guide: [Create an HTTP triggered function in Azure](https://docs.microsoft.com/en-us/azure/azure-functions/functions-create-first-function-python)
```
asavaritayal:
```suggestion
- [Azure Functions Python developer reference](https://aka.ms/python-functions-guide)  
```

Functions.Templates/Templates/HttpTrigger-Python/__init__.py
```diff
@@ -4,21 +4,24 @@
...
-    if not name:
-        try:
-            req_body = req.get_json()
-        except ValueError:
-            pass
-        else:
-            name = req_body.get('name')
+    logging.info("Python HTTP trigger function processed a request.")
+
+    # Get the value for `name` from URL arguments. `None` if not present.
```
asavaritayal:
```suggestion
    # Get the value of the `name` parameter from the query string of the URL.
```

Functions.Templates/Templates/HttpTrigger-Python/__init__.py
```diff
@@ -4,21 +4,24 @@
...
+    name = req.params.get("name")
+
+    # If `name` is not found in the URL arguments, check the JSON body if it exists.
```
asavaritayal:
```suggestion
    # If `name` is not found in the query string, check if it exists in the JSON body.
```

Functions.Templates/Templates/HttpTrigger-Python/__init__.py
```diff
@@ -4,21 +4,24 @@
 
 
 def main(req: func.HttpRequest) -> func.HttpResponse:
-    logging.info('Python HTTP trigger function processed a request.')
+    # Parse the request to the HTTP endpoint for a value corresponding to `name`.
```
asavaritayal:
```suggestion
    # Parse the HTTP request for the `name` parameter in the body or query string.
```

Functions.Templates/Templates/HttpTrigger-Python/README.md
```diff
@@ -0,0 +1,54 @@
...
+http://localhost:7071/api/<function-name>?name=Sophia
+```
+
+You should now get the response
+
+``` text
+Hello Sophia!
+```
+
+## Useful Resources
```
asavaritayal:
```suggestion
## Resources
```

Functions.Templates/Templates/HttpTrigger-Python/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+- Software: [Postman](https://www.getpostman.com/) is a cross-platform application for testing HTTP requests.
```
asavaritayal: This can be tricky wrt referencing third-party resources. Let's hold back for now. I'll follow up and confirm if we can add this to our templates. 

```suggestion

```

Functions.Templates/Templates/HttpTrigger-Python/README.md
```diff
@@ -0,0 +1,54 @@
...
+- Guide: [Create an HTTP triggered function in Azure](https://docs.microsoft.com/en-us/azure/azure-functions/functions-create-first-function-python)  
+- Examples: [Serverless Library - Python](https://serverlesslibrary.net/?technology=Functions%202.x&language=Python)
```
asavaritayal:
```suggestion
- [HTTP triggers and bindings](https://docs.microsoft.com/azure/azure-functions/functions-bindings-http-webhook)
- [Serverless Library](https://aka.ms/python-functions-samples)
```

.gitignore
t-eckert: Probably better to just include `.venv`. Avoids the potential of a name collision with `dev.env` and the like.

Functions.Templates/Templates/HttpTrigger-Python/README.md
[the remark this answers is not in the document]
t-eckert: I added it just because I have a programmer's penchant for lists that include themselves.

Functions.Templates/Templates/HttpTrigger-Python/README.md
t-eckert: Accepting your suggestion by changing my own code. Just wanted to keep the commit tree clean.

Functions.Templates/Templates/HttpTrigger-Python/__init__.py
t-eckert: Reopening. While testing this, I found that `hasattr(req, "get_json")` is always `True`, even when the request doesn't have JSON.

t-eckert:
Closing because of inactivity.

Issue #871, opened 2019-05-31
Timer - Template Improvements for Python


asavaritayal:
Child issue for https://github.com/Azure/azure-functions-templates/issues/869

- [ ] Complete the "TBDs" in the READMEs
- [ ] Provide more comprehensive code samples to better demonstrate the API surface. (HTTP template good but others like Blob and Queue are currently only single line logging statements.)
- [ ] Add comments for aka links to the binding reference articles. 
- [ ] Add `"scriptFile": "__init__.py", "entryPoint": "main"` to `function.json` for all templates.

PR #882, opened 2019-06-13
#871 python timer template

t-eckert:
This PR addresses issue #871 Timer - Template Improvements for Python.

- Code in `__init__.py` now imports `datetime` as `from datetime import datetime`. Smarter people than me disagree on which way to write it, but I tend to see more people preferring this.  
- `readme.md` -> `README.md`. More common.
- Removed `sample.dat` as sample input wasn't really appropriate for a timer trigger. 
- Renamed `mytimer` to `schedule`. I just like this name more. Plus, a cron job is more a scheduler than a timer (at least in my opinion).
- I am unsure if `schedule.past_due` will ever be `True`. I checked the documentation and while I can find it set to `False` in the class, I never see a check that might render it `True`, but I might be missing something.  
- Added documentation to README. There is room for more, but I wanted to get some work out for comments.


t-eckert:
Looks like the `sample.dat` is required to pass a test on Appveyor.

t-eckert:
Closing because of inactivity.

PR #890, opened 2019-07-05
Treat id for cosmos db document as key

maiqbal11:
Fixing default template experience for Python to be in line with published docs: https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-cosmosdb-v2#trigger---python-example.


maiqbal11:
@soninaren, Would like to get this into the current release of core tools since this is broken out-of-the-box for Python CosmosDB users.

maiqbal11:
@alrod, looks like @soninaren is OOF this week. Would like to get this into the core tools release for this host version.

PR #924, opened 2020-02-05
Exclude sampling of requests by default in host.json templates

soninaren:


soninaren:
related issue https://github.com/Azure/azure-functions-templates/issues/906

PR #927, opened 2020-02-15
Updating httpTrigger templates to provide a friendly message

### LINKED ISSUE #919: Update the default message for HTTP trigger template
Hey there Functions friends!

We (DevDiv) have been running user studies on the Azure Functions experience in VS Code over the last month. One consistent thing we see is that participants don't realize they have successfully created and ran the function app because they see an "error" the first time they run the app. On first launch, they see `Please pass a name on the query string or in the request body` and they immediately think something is wrong with the app.

Can this be updated to indicate the app is running? Something like

> "Your Azure Function app is running, pass a 'name' parameter to see it working."

I'm open to whatever, I'd just like to see this be a bit friendlier.

...and maybe get rid of `sample.dat` while you're in there? <3

soninaren:
resolves #919

### DIFF
--- a/Functions.Templates/Templates/HttpTrigger-CSharp/HttpTriggerCSharp.cs
+++ b/Functions.Templates/Templates/HttpTrigger-CSharp/HttpTriggerCSharp.cs
@@ -38,8 +38,8 @@ namespace Company.Function
             name = name ?? data?.name;
 
             return name != null
-                ? (ActionResult)new OkObjectResult($"Hello, {name}")
-                : new BadRequestObjectResult("Please pass a name on the query string or in the request body");
+                ? (ActionResult)new OkObjectResult($"Hello {name}, The HTTP trigger function executed succefully")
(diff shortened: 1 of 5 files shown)


Functions.Templates/Templates/HttpTrigger-CSharp/HttpTriggerCSharp.cs
```diff
@@ -38,8 +38,8 @@ public static class HttpTriggerCSharp
             name = name ?? data?.name;
 
             return name != null
-                ? (ActionResult)new OkObjectResult($"Hello, {name}")
-                : new BadRequestObjectResult("Please pass a name on the query string or in the request body");
+                ? (ActionResult)new OkObjectResult($"Hello {name}, The HTTP trigger function executed succefully")
```
mhoeger: nit: "this HTTP triggered function"?

PR #954, opened 2020-06-09
Explicitly including Functions json files

aaronpowell:
As noted in https://github.com/Azure/azure-functions-core-tools/issues/1963#issuecomment-641019816 because F# requires all files to be explicitly included with MSBuild, using `Update` rather than `Include` for the `host.json` and `local.settings.json` will result in them not being included and likely resulting in a failure to run `func start`.

This may also be related to https://github.com/Azure/azure-functions-durable-extension/issues/1140

### DIFF
--- a/Functions.Templates/ProjectTemplate/FSharp/Company.FunctionApp.fsproj
+++ b/Functions.Templates/ProjectTemplate/FSharp/Company.FunctionApp.fsproj
@@ -8,10 +8,10 @@
     <PackageReference Include="Microsoft.NET.Sdk.Functions" Version="1.0.36" />
   </ItemGroup>
   <ItemGroup>
-    <None Update="host.json">
+    <None Include="host.json">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
-    <None Update="local.settings.json">
+    <None Include="local.settings.json">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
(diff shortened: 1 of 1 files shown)


aaronpowell:
Use your power @cartermp 😉

aaronpowell:
@cartermp a comment on the issue thread https://github.com/Azure/azure-functions-core-tools/issues/1963#issuecomment-651818394 that puts a spanner in this. Do you think that there's a way with the project file to handle the include in a smarter way?

cartermp:
This would have to be in the F# compiler and/or build as tracked here: https://github.com/dotnet/fsharp/issues/8914

aaronpowell:
There's probably not a simple fix for it then. My only thought would be conditions on the include for debug only

johlrich:
It seems this PR wasn't merged because of concerns around accidental publishing the `local.settings.json` file. The `<CopyToPublishDirectory>Never</CopyToPublishDirectory>` already included in the template correctly prevents that file from appearing in the publish directory. 

I personally archive the _publish_ folder and use run from package and I've never seen local.settings.json included. 

Could another part of the tool using the _build_ output and not _publish_ output? If that's the case, any information leak is likely unrelated to this change.

jasiozet:
Has there any fix to this come since the closing?
I tried to setup azure function in F# and it remains an issue.

bigjonroberts:
This is really depressing to see core tooling like this left like this for so long.  It definitely feels like perfect has become the enemy of good here. 😞

Issue #971, opened 2020-08-08
Wrong parameter name in function.json for Python Event Hub Trigger


jimbobbennett:
When creating an event hub trigger in Python, the function that is created in the `__init__.py` file has a parameter called `events`:

```python
def main(events: List[func.EventHubEvent]):
```

In the `function.json` file where the binding is defined, the mapping is called `event`, not `events`:

```json
{
    "scriptFile": "__init__.py",
    "bindings": [
      {
        "type": "eventHubTrigger",
        "name": "event",
        "direction": "in",
        "eventHubName": "samples-workitems",
        "connection": "",
        "cardinality": "many",
        "consumerGroup": "$Default",
        "dataType": "binary"
      }
    ]
}
```

The `function.json` needs to be updated to change the `name` to `events`.

chadleong:
This has caused quite a headache for a beginner like myself. Took my awhile to figure out why my azure function isn't working.. should have checked the exception logs first thing but I was still figuring out the ApplicationInsights. This should be an urgent fix.

YaswantSaiKrishna:
Urgent fix required!!!

liliankasem:
Looks like this was fixed here https://github.com/Azure/azure-functions-templates/commit/949085f239ba6f457a6aef636cd2121ae905785c

PR #972, opened 2020-08-08
Fixed the parameter name for Python event hubs template

jimbobbennett:
Fix for #971.

The `__init__.py` file has a parameter called `events`, but the `function.json` file named the parameter `event`. This fixes the name in the json.


anthonychu:
Thanks @jimbobbennett. @soninaren can you please see if this can be merged?

PR #973, opened 2020-08-18
Add Durable Entities templates

bachuv:
Resolves [#1404](https://github.com/Azure/azure-functions-durable-extension/issues/1404). Adds Durable Entities templates for C#, JS, and TS.


Functions.Templates/Templates/DurableFunctionsEntityHttpStart-JavaScript/function.json
```diff
@@ -0,0 +1,25 @@
+﻿{
+  "bindings": [
+    {
+      "authLevel": "function",
+      "name": "req",
+      "type": "httpTrigger",
+      "direction": "in",
+      "route": "orchestrators/{functionName}",
```
anthonychu, line 8:
```suggestion
      "route": "counters/{id}",
```

Functions.Templates/Templates/DurableFunctionsEntityHttpStart-JavaScript/index.js
```diff
@@ -0,0 +1,11 @@
+﻿const df = require("durable-functions");
+
+module.exports = async function (context) {
+    const client = df.getClient(context);
+    const entityId = new df.EntityId("Counter", "myCounter");
```
anthonychu, line 5:
```suggestion
module.exports = async function (context, req) {
    const client = df.getClient(context);
    const entityId = new df.EntityId("Counter", req.params.id);
```

Functions.Templates/Templates/DurableFunctionsEntityHttpStart-JavaScript/index.js
```diff
@@ -0,0 +1,11 @@
...
+
+    await client.signalEntity(entityId, "add", 1);
+
+    const stateResponse = await client.readEntityState(entityId);
+    return stateResponse.entityState;
```
anthonychu, line 10: Having this line here implies that this would return the state after the preceding `signalEntity` has been completed. But there's no guarantee that this has happened. Suggest removing this.

Functions.Templates/Templates/DurableFunctionsEntity-JavaScript/metadata.json
```diff
@@ -0,0 +1,18 @@
+﻿{
+  "defaultFunctionName": "DurableFunctionsEntityJS",
```
anthonychu, line 2:
```suggestion
  "defaultFunctionName": "Counter",
```

Would it make more sense to default to `Counter`?

Functions.Templates/Templates/DurableFunctionsEntity-TypeScript/metadata.json
```diff
@@ -0,0 +1,18 @@
+﻿{
+  "defaultFunctionName": "DurableFunctionsEntityTS",
```
anthonychu, line 2:
```suggestion
  "defaultFunctionName": "Counter",
```

Would it make more sense to default to `Counter`?

Functions.Templates/Templates/DurableFunctionsEntityHttpStart-TypeScript/index.ts
```diff
@@ -0,0 +1,19 @@
+﻿﻿import { AzureFunction, Context, HttpRequest } from "@azure/functions"
+import * as df from "durable-functions"
+
+const httpStart: AzureFunction = async function (context: Context, req: HttpRequest) {
+    const client = df.getClient(context);
+    const id: string = context.bindingData.id;
```
anthonychu, line 6:
```suggestion
    const id: string = req.params.id;
```

... to match how we access the route param in [this template](https://github.com/Azure/azure-functions-templates/blob/dev/Functions.Templates/Templates/DurableFunctionsHttpStart-JavaScript/index.js#L5)

Functions.Templates/Templates/DurableFunctionsEntityHttpStart-TypeScript/index.ts
```diff
@@ -0,0 +1,19 @@
...
+    const entityId = new df.EntityId("Counter", id);
+
+    // reset value
+    await client.signalEntity(entityId, "reset", 1);
+
+    // increment value
+    await client.signalEntity(entityId, "add", 1);
+
+    // reads current state of entity
+    return await client.readEntityState<number>(entityId);
```
anthonychu, line 16: Wonder if it's better to leave these out, as it might lead someone to believe that these execute synchronously and the `readEntityState` reflects the result of the previous `signalEntity` calls.

Functions.Templates/Templates/DurableFunctionsEntityOrchestration-CSharp-2.x/DurableFunctionsEntityOrchestrationCSharp.cs
```diff
@@ -0,0 +1,57 @@
...
+            }
+        }
+
+        [FunctionName("DurableFunctionsEntityOrchestrationCSharp_CounterHttpStart")]
+        public static async Task<HttpResponseMessage> Run(
+        [HttpTrigger(AuthorizationLevel.Function)] HttpRequestMessage req,
+        [DurableClient] IDurableOrchestrationClient client,
+        ILogger log)
+        {
+            string instanceId = await client.StartNewAsync("DurableFunctionsEntityOrchestrationCSharp_CounterOrchestrator", null);
```
anthonychu, line 33: It might work better if this could be simplified to a single `signalEntity` call, without going through an orchestrator. This would allow us to use the template as-is in a quickstart.

Functions.Templates/Templates/DurableFunctionsEntityHttpStart-TypeScript/index.ts
bachuv: What would you recommend adding in place of these calls? Would it be better to just have one call to add to the `Counter` entity?

anthonychu: Actually, now that I think about it, can we do something like this?

```typescript
import { AzureFunction, Context, HttpRequest } from "@azure/functions"
import * as df from "durable-functions"

const httpStart: AzureFunction = async function (context: Context, req: HttpRequest) {
    const client = df.getClient(context);
    const id: string = context.bindingData.id;
    const entityId = new df.EntityId("Counter", id);

    if (req.method === "POST") {
         // increment value
        await client.signalEntity(entityId, "add", 1);
    } else {
        // read current state of entity
        return await client.readEntityState<number>(entityId);
    }
};

export default httpStart;
```

This should work quite well for a quickstart. @bachuv @cgillum any thoughts on this? If we're all in agreement, the JS and C# templates can do the same. Thanks.

cgillum: I like it! It is simple and avoids the problem you called out earlier.

bachuv: Looks good to me! I'll make the changes for JS and C# too.

davidmrdavid:
These look good to me, and I agree with Anthony's suggestions above. Thank you for these templates!

anthonychu:
@bachuv @cgillum checking in to see how things are going and if we could get this into the bundles release. /cc @AnatoliB

bachuv:
@anthonychu  I just pushed new updates. Let me know if you want me to make any more changes.

I also want to see if @soninaren has any comments about this PR so I added him as a reviewer.

cgillum:
FYI @bachuv it looks like there is a merge conflict that needs to be resolved.

Build/package-lock.json
```diff
@@ -3742,7 +3742,8 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true
+          "dev": true,
+          "optional": true
```
soninaren, line 3746: not sure why this file was changed. Can you revert this?

Functions.Templates/Bindings/bindings.json
```diff
@@ -1,1541 +1,1574 @@
...
+      ]
+    },
+    {
+      "type": "entityTrigger",
+      "displayName": "$entityTrigger_displayName",
+      "direction": "trigger",
+      "enabledInTryMode": false,
+      "documentation": "$content=Documentation\\entityTrigger.md",
+      "extension": {
+        "id": "Microsoft.Azure.WebJobs.Extensions.DurableTask", "version": "2.2.2"
```
soninaren, line 1164: This should be `2.3.0`

Functions.Templates/Resources/Resources.resx
```diff
@@ -1804,4 +1804,28 @@ function app.</value>
...
+    <value>(Optional) Name of your Entity. Defaults to the name of you Entity if not specified.</value>
+  </data>
+  <data name="entityTrigger_entity_label" xml:space="preserve">
+    <value>Entity Name</value>
+  </data>
+  <data name="entityTrigger_name_help" xml:space="preserve">
+    <value>The name used to identify this binding in your code</value>
+  </data>
+  <data name="entityTrigger_name_label" xml:space="preserve">
+    <value>entityTrigger_name_label</value>
```
soninaren, line 1829: This should be `Name`

Functions.Templates/Templates/DurableFunctionsEntity-JavaScript/metadata.json
soninaren: Not sure if this was resolved.

Functions.Templates/Templates/DurableFunctionsEntityHttpStart-CSharp-2.x/DurableFunctionsEntityHttpStartCSharp.csx
```diff
@@ -0,0 +1,22 @@
+#r "Microsoft.Azure.WebJobs.Extensions.DurableTask"
+#r "Newtonsoft.Json"
+
+using System.Net;
+using Microsoft.Azure.WebJobs.Extensions.DurableTask;
+
+public static async Task<HttpResponseMessage> Run(
+    [HttpTrigger(AuthorizationLevel.Function, Route = "Counter/{entityKey}")] HttpRequestMessage req,
+    [DurableClient] IDurableEntityClient client,
```
soninaren, line 9: This was throw compile error. The attributes are not needed in csx file.
```c#
[HttpTrigger(AuthorizationLevel.Function, Route = "Counter/{entityKey}")] HttpRequestMessage req, [DurableClient] IDurableEntityClient client
```
Was this tested locally or on portal?

Functions.Templates/Templates/DurableFunctionsEntityOrchestration-CSharp-2.x/.template.config/template.json
```diff
@@ -0,0 +1,59 @@
...
+  "postActions": [
+    {
+      "Description": "Adding Reference to Microsoft.Azure.WebJobs.Extensions.DurableTask NuGet package",
+      "ActionId": "B17581D1-C5C9-4489-8F0A-004BE667B814",
+      "ContinueOnError": "true",
+      "ManualInstructions": [],
+      "args": {
+        "referenceType": "package",
+        "reference": "Microsoft.Azure.WebJobs.Extensions.DurableTask",
+        "version": "2.2.2",
```
soninaren, line 38: this should be `2.3.0`

Functions.Templates/Templates/DurableFunctionsEntityOrchestration-CSharp-2.x/DurableFunctionsEntityOrchestrationCSharp.cs
```diff
@@ -0,0 +1,48 @@
...
+using Microsoft.Azure.WebJobs;
+using Microsoft.Azure.WebJobs.Extensions.DurableTask;
+using Microsoft.Azure.WebJobs.Extensions.Http;
+using Newtonsoft.Json.Linq;
+
+namespace Company.Function
+{
+    public static class DurableFunctionsEntityOrchestrationCSharp
+    {
+        [FunctionName("DurableFunctionsEntityOrchestrationCSharp_CounterHttpStart")]
```
soninaren, line 13: `DurableFunctionsEntityOrchestrationCSharp` would be replaced by the function name mentioned by the customer. Not sure if that would break the durable orchestration logic.

soninaren (changes requested):
Looks good mostly. Can you make sure to test this end to end? On portal and core tools..

Functions.Templates/Templates/DurableFunctionsEntityOrchestration-CSharp-2.x/DurableFunctionsEntityOrchestrationCSharp.cs
bachuv: It should be fine.

soninaren (approved):
Would need to rebase on dev to merge.

PR #977, opened 2020-08-27
Update the initial __init__.py file so users can quickly see what needs to be changed/updated to manipulate their function

ijoosong:
- Added comments.  I think that having comments will alleviate some of the questions that a user has when they first open the __init__.py file
 - Removed the if/try/else logic because that assumes user is trying to create a POST request.  Most of the quickstarts/tutorials assume the user is going to hit it with a web browser or at least as a GET request.

### DIFF
--- a/Functions.Templates/Templates/HttpTrigger-Python/__init__.py
+++ b/Functions.Templates/Templates/HttpTrigger-Python/__init__.py
@@ -1,20 +1,22 @@
 import logging
 
+# Add any dependencies you might need here
 import azure.functions as func
 
 
 def main(req: func.HttpRequest) -> func.HttpResponse:
+    """
+    This is the function that will be called when calling your Azure Function.
+    Currently, it is a template for an HTTP Trigger that says hello when calling your URL.
+    Try changing the code found below to manipulate your function!
+    If you want to change where your Azure Function looks, take a look at 'function.json'
+    """
     logging.info('Python HTTP trigger function processed a request.')
 
+    # Getting the 'name' parameter from the url string
     name = req.params.get('name')
-    if not name:
-        try:
-            req_body = req.get_json()
-        except ValueError:
-            pass
-        else:
-            name = req_body.get('name')
 
+    # Quick check to see if 'name' parameter exists from above
     if name:
         return func.HttpResponse(f"Hello, {name}. This HTTP triggered function executed successfully.")
     else:


Functions.Templates/Templates/HttpTrigger-Python/__init__.py
```diff
@@ -1,20 +1,22 @@
...
+    This is the function that will be called when calling your Azure Function.
+    Currently, it is a template for an HTTP Trigger that says hello when calling your URL.
+    Try changing the code found below to manipulate your function!
+    If you want to change where your Azure Function looks, take a look at 'function.json'
+    """
     logging.info('Python HTTP trigger function processed a request.')
 
+    # Getting the 'name' parameter from the url string
     name = req.params.get('name')
-    if not name:
```
ijoosong, line 10: There are no quickstarts or tutorials that start with using a POST request with a body.  I don't see why we need this here to further complicate something when users are first starting out.

vrdmr: I partially agree - this snippet does help folks to know how to read the body. I think it'll be better to add methods in the function.json (get and post) to fix this issue.

Functions.Templates/Templates/HttpTrigger-Python/__init__.py
```diff
@@ -1,20 +1,22 @@
...
 import azure.functions as func
 
 
 def main(req: func.HttpRequest) -> func.HttpResponse:
+    """
+    This is the function that will be called when calling your Azure Function.
+    Currently, it is a template for an HTTP Trigger that says hello when calling your URL.
+    Try changing the code found below to manipulate your function!
+    If you want to change where your Azure Function looks, take a look at 'function.json'
+    """
```
vrdmr, line 13: Could you explicitly mention `scriptFile` in the comment?

Functions.Templates/Templates/HttpTrigger-Python/__init__.py
```diff
@@ -1,20 +1,22 @@
 import logging
 
+# Add any dependencies you might need here
```
Hazhzeng, line 3: Dependencies need to be defined in requirements.txt

Functions.Templates/Templates/HttpTrigger-Python/__init__.py
ijoosong: So the function.json comes with GET and POST today.  Are you saying that we fix that to just GET? 

Another solution to this could be to have helper functions that we've moved out of this main function dealing with GET and POST requests.  If we want to keep it in, we should be very explicit about what it is and why it's there.  That would be more code but it could help users see what's happening.

Functions.Templates/Templates/HttpTrigger-Python/__init__.py
ijoosong: You're right. I'll update this to be more explicit

Functions.Templates/Templates/HttpTrigger-Python/__init__.py
Hazhzeng: Compared with the JavaScript HTTP trigger template, https://github.com/Azure/azure-functions-templates/blob/dev/Functions.Templates/Templates/HttpTrigger-JavaScript/index.js#L4

The `name` variable can be assigned from query parameter (GET) or request body (POST).
```javascript
const name = (req.query.name || (req.body && req.body.name));
```
Here, the Python template is just following the same pattern.

Hazhzeng: I don't recommend removing this.

Functions.Templates/Templates/HttpTrigger-Python/__init__.py
ijoosong: Varad, what do you mean by `scriptFile` here?

liliankasem (changes requested):
Looks like there are still pending code review comments that need to be addressed

soninaren:
Closing as stale PR

PR #980, opened 2020-09-14
Changes for Python GA release

davidmrdavid:
**WIP**: _Do not merge yet, this is a draft PR..._ 

This PR makes some adjustements necessary for the announcement of Durable Python GA.
Changes so far:
* Removed `(preview)` labels on activity, orchestrator, and http starter templates
* Included new Counter Entity template for Python

A few notes:
1) This PR is currently dependent on the changes introduced by @bachuv's currently unmerged PR here: https://github.com/Azure/azure-functions-templates/pull/973 . As a result ,we shouldn't merge this until those changes are merged first. I'm currently avoiding pulling those changes here as I'm unsure about whether or not that PR is stable yet.
2) I see there are a handful of templates for Python with the suffix `-1.x`. For instance: `DurableFunctionsOrchestrator-Python-1.x`. I assume that those mean templates for Functions V1. @anthonychu , should we remove those? By the same token, I did not make changes to `-1.x`-suffixed directories, under the expectation that I'll be deleting them, but I'll wait the clarification first.


Build/PackageFiles/ExtensionBundleTemplates-2.x.nuspec
```diff
@@ -156,7 +156,10 @@
     <file src="../../Functions.Templates/templates/DurableFunctionsHttpStart-Python-2.x/metadata.json" target="Templates/DurableFunctionsHttpStart-Python/metadata.json" />
     <file src="../../Functions.Templates/templates/DurableFunctionsHttpStart-TypeScript/function.json" target="templates/DurableFunctionsHttpStart-TypeScript/function.json" />
     <file src="../../Functions.Templates/templates/DurableFunctionsHttpStart-TypeScript/metadata.json" target="templates/DurableFunctionsHttpStart-TypeScript/metadata.json" />
-    <file src="../../Functions.Templates/templates/DurableFunctionsHttpStart-TypeScript/index.ts" target="templates/DurableFunctionsHttpStart-TypeScript/index.ts" />    
+    <file src="../../Functions.Templates/templates/DurableFunctionsHttpStart-TypeScript/index.ts" target="templates/DurableFunctionsHttpStart-TypeScript/index.ts" />
```
davidmrdavid, line 159: not sure how this line got added....I'll remove it

Functions.Templates/Templates/DurableFunctionsEntity-Python/metadata.json
```diff
@@ -0,0 +1,19 @@
...
+    "$temp_category_durableFunctions"
+  ],
+  "categoryStyle": "other",
+  "enabledInTryMode": false,
+  "userPrompt": [],
+  "extensions": [
+    {
+      "id": "Microsoft.Azure.WebJobs.Extensions.DurableTask",
+      "version": "2.3.0"
+    }
```
davidmrdavid, line 17: Honestly unsure about what this block achieves, but I got it from @bachuv 's PR. Can anyone clarify what it does?

anthonychu: I think this is used in some places where it automatically does `func extensions install` for you when you add the binding to a function (like the portal?). Not sure if this would be used anywhere for Python but @soninaren will know.

anthonychu:
@davidmrdavid I'm good with removing templates for v1. We'll need to update docs and quickstarts with instructions on how to update host.json.

@soninaren We want to require Durable Functions extension v2 (bundles v2) for Python Durable. This means the templates only appear when bundles v2 is selected. How does this work for people who are using extensions.json?

There's going to be some confusion here. @davidmrdavid @cgillum any ideas on how to improve discoverability? Projects are still defaulting to v1 and the templates will not be there unless user manually updates host.json to bundles v2.

We could do some things in tooling:
- Default to v2 https://github.com/Azure/azure-functions-extension-bundles/issues/35
- Command to switch between versions https://github.com/Azure/azure-functions-core-tools/issues/1913
- VS Code gesture to switch between versions https://github.com/microsoft/vscode-azurefunctions/issues/2058

cgillum:
Yeah, we should definitely be defaulting to v2 for new apps. As far as tooling to help folks migrate, I see the feedback that the VS Code folks provided and it makes sense to me that maybe we should just have customers make the update themselves.  But since we already have customers using v1 bundles, I think we may need to continue to support v1 bundles, but log a warning message that tells the user to update to the v2 bundles. We discussed in our sync maybe throwing an error, but I'm starting to think that might not be such a good idea.

davidmrdavid:
@anthonychu 
Another idea would be to have VSCode raise a notification reminding / suggests users that they upgrade their bundles. I believe this is different from the gesture you suggest, as this would be more proactive in getting users to convert to V2 instead of requiring that they know about some gesture.

@cgillum @anthonychu 
If the warning is "loud" enough, I would be ok with letting customers continue using V1. Another approach would be to error-out only when a user tries to combine the upcoming _stable_ / GA release with V1, but continue to support V1 + _beta_ releases of Durable Python with just a warning. That would simply require that the SDK send some extra metadata about its version to the extension. What do y'all think

cgillum:
@davidmrdavid I guess if I step back and think about it, customers would run into this only after the upgrade to our GA Python SDK. I suppose at that point they're already upgrading something, so maybe we shouldn't feel too bad about asking them to upgrade one more thing (the bundles version). So I think I'm on board with your suggestion. :)

Having VS Code recommend the upgrade sounds like a good compromise to me. It might be more convincing if there is some prior art though, like if VS Code has a way of notifying about out-of-date npm dependencies.

Build/package-lock.json
```diff
@@ -3742,7 +3742,8 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true
+          "dev": true,
+          "optional": true
```
davidmrdavid, line 3746: I think I wasn't supposed to commit this file...

davidmrdavid:
Hi @soninaren , I believe this PR is almost ready for merge. I just need to go through some testing to make sure it's working as expected, which I'll be doing tomorrow. In the meantime, could give this a look and make sure nothing looks out of place? Thank you!

@anthonychu just FYI

Build/package-lock.json
soninaren: Yes, please remove the changes to this file.

Functions.Templates/Templates/DurableFunctionsEntity-Python/metadata.json
```diff
@@ -0,0 +1,19 @@
+{
```
soninaren, line 1: Where is this template used. I don't see it in any .nuspec file

Functions.Templates/Templates/DurableFunctionsEntity-Python/__init__.py
```diff
@@ -0,0 +1,23 @@
+import logging
```
soninaren, line 1: Same as below. Where is this template used. I don't see it in any .nuspec file

soninaren (changes requested):

Functions.Templates/Templates/DurableFunctionsEntity-Python/function.json
```diff
@@ -1,9 +1,9 @@
-﻿{
+{
   "scriptFile": "__init__.py",
   "bindings": [
     {
       "name": "context",
-      "type": "orchestrationTrigger",
+      "type": "entityTrigger",
```
soninaren, line 6: This file needs to go in 2.x bundle, the path folder needs to be renamed `-2.x`

Functions.Templates/Templates/DurableFunctionsEntity-Python/metadata.json
soninaren, line 1: Should this be in 1.x bundle or 2.x or both?

davidmrdavid: only 2.x

soninaren (approved):

PR #989, opened 2020-10-27
Fixed gulp-build error due to faulty path in nuspec file

davidmrdavid:
Somehow the CI did not catch this while merging my previous PR, but apparently there was a missing `-2.x` suffix in the nuspec file for Durable Functions Python


Francisco-Gamino (approved):
LGTM

PR #993, opened 2020-11-03
Updated durable functions templates in C# to use nameof 

lbugnion:
This uses nameof wherever possible instead of magic strings. This makes navigating to the symbols (F12 in Visual Studio) much easier and more convenient. It also avoids errors when renaming a method.

### DIFF
--- a/Functions.Templates/Templates/DurableFunctionsEntityHttp-CSharp-2.x/DurableFunctionsEntityHttpCSharp.cs
+++ b/Functions.Templates/Templates/DurableFunctionsEntityHttp-CSharp-2.x/DurableFunctionsEntityHttpCSharp.cs
@@ -16,7 +16,7 @@ namespace Company.Function
         [DurableClient] IDurableEntityClient client,
         string entityKey)
         {
-            var entityId = new EntityId("Counter", entityKey);
+            var entityId = new EntityId(nameof(Counter), entityKey);
 
             if (req.Method == HttpMethod.Post)
             {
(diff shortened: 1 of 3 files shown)


anthonychu:
Thanks @lbugnion.

cc @cgillum @ConnorMcMahon

liliankasem (approved):

davidmrdavid:
Thanks for your contribution. This seems right to me but, before merging, I think it'd be good to check-in with @AnatoliB and @cgillum as to whether or not we want to incorporate this pattern into the templates. If possible, I'd like to wait until they can give this an approval themselves :)

liliankasem:
> Thanks for your contribution. This seems right to me but, before merging, I think it'd be good to check-in with @AnatoliB and @cgillum as to whether or not we want to incorporate this pattern into the templates. If possible, I'd like to wait until they can give this an approval themselves :)

@davidmrdavid @AnatoliB  @cgillum - any thoughts on this before we merge?

davidmrdavid:
My advice still stands: I like the change, but I want to get at least one more DF person to approve since this affects our quickstart experience. I'll wait for either of the tagged folks above to chime in :-)

AnatoliB (approved):
I like this change.

cgillum (approved):
LGTM! This is definitely a best practice, so having it in our templates feels like the right thing to do.

davidmrdavid:
@liliankasem go ahead and merge then!

PR #1010, opened 2021-01-21
Use safer binding names for Durable Python in V1

davidmrdavid:
This PR removes and fixes incorrect template binding names in V1 for Durable Python.


ConnorMcMahon (approved):
LGTM!

PR #1015, opened 2021-01-28
Add net5 initial templates

ankitkumarr:
I am not sure if we need the images for vs17 as net5 wouldn't really work there.


Functions.Templates/Functions.Templates.csproj
```diff
@@ -105,6 +105,7 @@
     <Content Include="ProjectTemplate\local.settings.json" />
     <Content Include="ProjectTemplate\_gitignore" />
     <Content Include="Templates\BlobTrigger-CSharp\BlobTriggerCSharp.cs" />
+    <Content Include="Templates\BlobTrigger-CSharp-Isolated\BlobTriggerCSharp.cs" />
```
ankitkumarr, line 108: I may not need to update this file actually.

Functions.Templates/ProjectTemplate/CSharp-Isolated/.template.config/template.json
```diff
@@ -0,0 +1,56 @@
...
+    "tags": {
+        "language": "C#",
+        "type": "project"
+    },
+    "sourceName": "Company.FunctionApp",
+    "symbols": {
+        "Namespace": {
+            "description": "namespace for the generated code",
+            "replaces": "Company.FunctionApp",
+            "defaultValue": "FunctionApp",
```
ankitkumarr, line 22: Need to verify if we can replace `Company.FunctionApp` with directory name. This is for the namespace in `Program.cs`.

Build/PackageFiles/ItemTemplates-Isolated.nuspec
```diff
@@ -0,0 +1,64 @@
+<?xml version="1.0"?>
+<package >
+  <metadata>
+    <id>Microsoft.Azure.Functions.Worker.ItemTemplates</id>
```
ankitkumarr, line 4: The new package ID for the isolated item templates

Build/PackageFiles/ItemTemplates-Isolated.nuspec
```diff
@@ -0,0 +1,64 @@
...
+    <version>1.0.0</version>
+    <authors>Microsoft</authors>
+    <licenseUrl>https://go.microsoft.com/fwlink/?linkid=2028464</licenseUrl>
+    <repository type="git" url="https://github.com/Azure/azure-functions-templates" />
+    <projectUrl>https://go.microsoft.com/fwlink/?linkid=848145</projectUrl>
+    <copyright>© Microsoft Corporation. All rights reserved.</copyright>
+    <requireLicenseAcceptance>true</requireLicenseAcceptance>
+    <description>Collection of templates for Azure Functions</description>
```
soninaren, line 12: Can you update the description to help differentiate between isolated and in-proc packages?

Build/PackageFiles/ProjectTemplates-Isolated.nuspec
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="utf-8"?>
```
soninaren, line 1: Are you planning to add F# project template as well?

Functions.Templates/ProjectTemplate/CSharp-Isolated/.template.config/dotnetcli.host.json
```diff
@@ -0,0 +1,3 @@
+{
+
```
soninaren, line 2: Don't need this file. Can delete it.

Functions.Templates/ProjectTemplate/CSharp-Isolated/.template.config/template.json
soninaren: Use this file as an example. https://github.com/Azure/azure-functions-templates/blob/dev/Functions.Templates/Templates/CosmosDBTrigger-CSharp/.template.config/template.json. Match the sourceName in respective locations. It should be same as the class file name. Cannot keep class file name as program.cs otherwise all `.cs` files for function would end up being named program.cs

Functions.Templates/ProjectTemplate/CSharp-Isolated/Program.cs
```diff
@@ -0,0 +1,26 @@
+using System.Threading.Tasks;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Azure.Functions.Worker.Configuration;
+
+namespace Company.FunctionApp
+{
+    class Program
```
soninaren, line 8: Use this file as an example. Replace program with SourceName value. Match the sourceName in respective locations. https://github.com/Azure/azure-functions-templates/blob/dev/Functions.Templates/Templates/CosmosDBTrigger-CSharp/CosmosDBTriggerCSharp.cs

soninaren (changes requested):

Build/PackageFiles/ProjectTemplates-Isolated.nuspec
ankitkumarr: For now, we aren't planning on. We may later. But this release won't include them.

Functions.Templates/ProjectTemplate/CSharp-Isolated/Program.cs
ankitkumarr: This would still be `Program`. This is the main file (new to this model).

phenning (approved):
@ankitkumarr I took another look at the proposed feed updates in the doc. Given that, these changes look good.

Functions.Templates/ProjectTemplate/CSharp-Isolated/.template.config/template.json
```diff
@@ -0,0 +1,55 @@
...
+    "author": "Microsoft",
+    "classifications": [
+        "Azure Functions",
+        "Solution"
+    ],
+    "name": "Azure Functions",
+    "generatorVersions": "[1.0.0.0-*)",
+    "groupIdentity": "Microsoft.AzureFunctions.ProjectTemplates",
+    "precedence": "100",
+    "identity": "Microsoft.AzureFunctions.ProjectTemplate.CSharp.Isolated.2.x",
```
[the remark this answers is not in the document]
ejizba: Yeah we want this to be 3.x to work in VS Code

There's an [old issue](https://github.com/Azure/azure-functions-templates/issues/916) about whether or not _item_ templates should use 3.x (and we should probably keep that a separate discussion), but the _project_ template has been 3.x for a while even in non-isolated.

Functions.Templates/Templates/HttpTrigger-CSharp-Isolated/HttpTriggerCSharp.cs
```diff
@@ -0,0 +1,31 @@
...
+        [FunctionName("HttpTriggerCSharp")]
+        public static HttpResponseData Run([HttpTrigger(AuthorizationLevel.AuthLevelValue, "get", "post")] HttpRequestData req,
+            FunctionExecutionContext executionContext)
+        {
+            var logger = executionContext.Logger;
+            logger.LogInformation("C# HTTP trigger function processed a request.");
+
+            var response = new HttpResponseData(HttpStatusCode.OK);
+            var headers = new Dictionary<string, string>();
+            headers.Add("Date", "Mon, 18 Jul 2016 16:06:00 GMT");
```
ejizba, line 22: Do we really need to hard-code a date like this? Seems a bit odd for the default template

ejizba (approved):
Seems to be working for me with VS Code 👍🎉

Functions.Templates/ProjectTemplate/CSharp-Isolated/.template.config/template.json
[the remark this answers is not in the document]
soninaren: It should be 3.x. Should we do the same for item templates as well? For item templates i believe we had to revert to 2.x because VS was throwing null reference exception.

Functions.Templates/Templates/BlobTrigger-CSharp-Isolated/.template.config/template.json
```diff
@@ -0,0 +1,71 @@
+{
+    "author": "Microsoft",
+    "classifications": [
+        "Azure Function",
+        "Trigger",
+        "Blob"
+    ],
+    "name": "BlobTrigger",
+    "identity": "Azure.Function.CSharp.Isolated.BlobTrigger.2.x",
```
soninaren, line 9: Change to 3.x

Functions.Templates/Templates/BlobTrigger-CSharp-Isolated/metadata.json
```diff
@@ -0,0 +1,21 @@
+{
+    "defaultFunctionName": "BlobTrigger",
```
soninaren, line 2: Don't need this file if we don't have .csx templates

Functions.Templates/Templates/CosmosDBTrigger-CSharp-Isolated/metadata.json
```diff
@@ -0,0 +1,24 @@
+{
```
soninaren, line 1: remove all metadata.json files

soninaren (changes requested):

Functions.Templates/ProjectTemplate/CSharp-Isolated/.template.config/template.json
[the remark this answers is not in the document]
ankitkumarr: Yeah I think I may accidentally used the `dev` branch value instead of `v3.x`. I will update this Project Template value to be `3.x` just to align with what we have. I will leave item templates as is, and we can take that issue separately if we do that across .NET Core 3 as well.

Functions.Templates/Templates/HttpTrigger-CSharp-Isolated/HttpTriggerCSharp.cs
ankitkumarr: This entire template code will actually change in a week or so. I just put what we had in samples/ as placeholder. But we are actively updating that. As soon as we release Nugets and updated samples this week, I will update the contents here.

Functions.Templates/Templates/BlobTrigger-CSharp-Isolated/.template.config/template.json
ankitkumarr: Let's leave this to `2.x` to match what we do in .NET Core 3? When we update there, we can update these too.

Functions.Templates/ProjectTemplate/CSharp-Isolated/.template.config/template.json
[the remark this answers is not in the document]
soninaren: We cannot do that for .Net Core 3 because of legacy code in old VS versions. In fact we did that and we had to revert it because of issues. If we do that here too then we would be stuck with 2.x in item templates for .Net 5 worker. Also there is no reason for us not use 3.x in item templates given that they are completely new and in a separate package.

Functions.Templates/Templates/BlobTrigger-CSharp-Isolated/.template.config/template.json
soninaren: Can't change to 3.x in .NET Core 3, if we do that VS would throw null reference exception. We should change these while we have the chance :)

ankitkumarr: Fair! I changed them.

soninaren (approved):

PR #1024, opened 2021-02-25
[dotnet-isolated] Type update, and version bump

ankitkumarr:


fabiocav (approved):

Functions.Templates/Templates/TimerTrigger-CSharp-Isolated/TimerTriggerCSharp.cs
```diff
@@ -11,7 +11,7 @@ public static void Run([TimerTrigger("ScheduleValue")] MyInfo myTimer, FunctionC
         {
             var logger = context.GetLogger("TimerTriggerCSharp");
             logger.LogInformation($"C# Timer trigger function executed at: {DateTime.Now}");
-            logger.LogInformation($"Next timer schedule at: {timerInfo.ScheduleStatus.Next}");
+            logger.LogInformation($"Next timer schedule at: {myTimer.ScheduleStatus.Next}");
```
brettsam, line 14: this doesn't work yet, right? i have the changes locally i think...

ankitkumarr: As discussed offline (well, online but different channel), we are referencing a POCO here, so this works for now. We may want to update that with our own type later.

PR #1029, opened 2021-03-08
Remove duplicate Python Durable Orchestrator

### LINKED ISSUE #1028: Durable orchestration appears twice in 2.x Python templates
image

Looks like it's because it is included twice:

https://github.com/Azure/azure-functions-templates/blob/5fe6dae42a2048794c89b62799e9892bd0ff7c7a/Build/PackageFiles/ExtensionBundleTemplates-2.x.nuspec#L203-L205

https://github.com/Azure/azure-functions-templates/blob/5fe6dae42a2048794c89b62799e9892bd0ff7c7a/Build/PackageFiles/ExtensionBundleTemplates-2.x.nuspec#L224-L226

/cc @davidmrdavid

anthonychu:
Resolves #1028

/cc @davidmrdavid


davidmrdavid (commented):
Thanks for this! There does seem to be a duplicate definition here, but I personally don't see the duplicate orchestrator problem myself when attempting to reproduce locally. However, some folks have reported seeing this issue temporarily and then seen it get fixed after a bit. I wonder if it's some caching issue exacerbated, but not caused by, that's by the duplicate declaration.

All that being said, let wait on this merge just until we're past GA. That will also give me time to test this locally. Thank you :)

anthonychu:
Make sure you change bundles to v2 and reload vs code. Ok with waiting on this since we expect most to use v1.

davidmrdavid:
Ok great! And yes, I am still unable to reproduce this even after reloading VSCode and using bundles V2. Regardless, we'll almost certainly merge this asap

davidmrdavid (approved):
Just tested it: LGTM!

davidmrdavid:
@soninaren, can I go ahead and merge?

soninaren:
> @soninaren, can I go ahead and merge?

yes

PR #1048, opened 2021-04-26
updating project templates for .net 6

brettsam:


Functions.Templates/ProjectTemplate/CSharp-Isolated/Company.FunctionApp.csproj
```diff
@@ -1,6 +1,6 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>net5.0</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
```
ankitkumarr, line 3: hmm I don't know if this needs to change necessarily.
For v4 Functions, are we thinking that the default for .NET Isolated is .NET 6.0? If so, then this makes sense.

ankitkumarr (approved):
Looks fine to me, assuming that this will be in a V4 branch as you comment said.
And would be good to confirm that we are changing the default .NET isolated target. As I think that will have some tooling implications.

brettsam:
This went in elsewhere...

PR #1065, opened 2021-06-15
[v2.x] Update Kafka Extension related %Password% AppSettings 

TsuyoshiUshio:
I create the Kafka extension template, however, the template include an issue. 
We use `%Password%` to set the password, and fetch it from AppSettings, however, it doesn't work on Azure. 
Probably, it is reserved word. 

I switch it into `%EventHubsConnectionString%` since the broker template is written for EventHubs Kafka API. 
I'll Create v3 PR as well. 

CC: @pragnagopa

### DIFF
--- a/Functions.Templates/Bindings/bindings.json
+++ b/Functions.Templates/Bindings/bindings.json
@@ -209,7 +209,7 @@
         {
           "name": "password",
           "value": "string",
-          "defaultValue": "%EventHubsConnectionString%",           
+          "defaultValue": "%KafkaPassword%",           
           "resource": "AppSetting",
           "required": false,
           "label": "$kafka_password_label",
@@ -337,7 +337,7 @@
         {
           "name": "password",
           "value": "string",
-          "defaultValue": "%EventHubsConnectionString%",           
+          "defaultValue": "%KafkaPassword%",           
           "resource": "AppSetting",
           "required": false,
           "label": "$kafka_password_label",
--- a/Functions.Templates/Bindings/bindings.json
+++ b/Functions.Templates/Bindings/bindings.json
@@ -164,7 +164,7 @@
       "enabledInTryMode": true,
       "documentation": "$content=Documentation\\kafkaTrigger.md",
       "extension": {
-        "id": "Microsoft.Azure.WebJobs.Extensions.Kafka", "version": "3.2.1"
+        "id": "Microsoft.Azure.WebJobs.Extensions.Kafka", "version": "3.3.2"
       },
       "settings": [
         {
@@ -292,7 +292,7 @@
       "enabledInTryMode": false,
       "documentation": "$content=Documentation\\kafkaOut.md",
       "extension": {
-        "id": "Microsoft.Azure.WebJobs.Extensions.Kafka", "version": "3.2.1"
+        "id": "Microsoft.Azure.WebJobs.Extensions.Kafka", "version": "3.3.2"
       },
       "settings": [
         {
--- a/Functions.Templates/Bindings/bindings.json
+++ b/Functions.Templates/Bindings/bindings.json
@@ -209,7 +209,7 @@
         {
           "name": "password",
           "value": "string",
-          "defaultValue": "%Password%",           
+          "defaultValue": "%EventHubsConnectionString%",           
           "resource": "AppSetting",
           "required": false,
           "label": "$kafka_password_label",
@@ -337,7 +337,7 @@
         {
           "name": "password",
           "value": "string",
-          "defaultValue": "%Password%",           
+          "defaultValue": "%EventHubsConnectionString%",           
           "resource": "AppSetting",
           "required": false,
           "label": "$kafka_password_label",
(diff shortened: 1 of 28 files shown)


Functions.Templates/Bindings/bindings.json
```diff
@@ -209,7 +209,7 @@
         {
           "name": "password",
```
pragnagopa, line 210: rename name as well?

Functions.Templates/Templates/KafkaOutput-CSharp/.template.config/template.json
```diff
@@ -41,8 +41,8 @@
         "Password": {
             "description": "SASL password for use with the PLAIN and SASL_SCRAM",
```
pragnagopa, line 42: Update description to indicate this is a connection string

Functions.Templates/Templates/KafkaOutput-CSharp/KafkaOutputCSharp.cs
```diff
@@ -18,7 +18,7 @@ public class KafkaOutputCSharp
             [Kafka("BrokerList",
                    "topic",
                    Username = "$ConnectionString",
```
pragnagopa, line 20: Connection string will include all the information needed to connect right? do you still need username?

Functions.Templates/Templates/KafkaOutput-JavaScript/function.json
```diff
@@ -16,7 +16,7 @@
             "brokerList" : "BrokerList",
             "topic" : "topic",
             "username" : "$ConnectionString",
-            "password" : "%Password%",
+            "password" : "%EventHubsConnectionString%",
```
pragnagopa, line 19: please check if you can remove username and password and only have connection string

pragnagopa (changes requested):
Added few comments

Functions.Templates/Bindings/bindings.json
TsuyoshiUshio: Hi @pragnagopa 

Thank you for the review. Probably it is ok for password.   
Let me explain the context. (it is bit confusing) 

The parameter that I want to configure is `password` it is credential for kafka. usually, it used with `username` and `password`  It is used for kafka broker like Confluent cloud. 
Event Hubs has a kafka compatible API. EventHubs only have connection string. If a kakfa client want to connect to EventHubs with Kafka API, it will set `username` as `$ConnectionString` (fixed value)  and `password` as connection string.  So I use `%Password%` as a placefolder. However, it doesn't work on Azure. Probably, reserved word. So that I change the placefolder from `%Password%` to  `%EventHubsConnectionString%` .

Functions.Templates/Templates/KafkaOutput-CSharp/.template.config/template.json
TsuyoshiUshio: As I explained last comment, it is exactly `password` and explanation is fine. Only missing point will be , in case of EventHubs, we set Connection String. I explain it on the Readme.md part. Is it enough or should we add it in here as well?

Functions.Templates/Templates/KafkaOutput-CSharp/KafkaOutputCSharp.cs
TsuyoshiUshio: Yes. it is the rule of the Kafka client connect to EventHubs.

Functions.Templates/Templates/KafkaOutput-CSharp/.template.config/template.json
pragnagopa: Please update the description to include for EventHubAPI - its connection string

Functions.Templates/Templates/KafkaOutput-JavaScript/function.json
TsuyoshiUshio: Hi @pragnagopa 

We can't remove the username and password, since it is designed for KafkaAPI. 
`ConnectionString` is used for EventHubs only.  Other kafka usage uses user/password (or credentials) 
So that the parameter name should be `username` and `password`  I understand people confuse the naming. 

I just want to avoid `%Password%` since it looks reserved.  How about change it form `%EventHubsConnectionString%` to `%KafkaPassword%`?  Do you think it is less confusing?

TsuyoshiUshio:
Hi @pragnagopa  I update the name from `EventHubsConnectionString` to `KafkaPassword`. 
I also tested on Azure if the `KafkaPassword` is available. It works.

TsuyoshiUshio:
Hi @soninaren 

The CI suddenly fails, it looks the issue of the NuGet feed authentication.  I haven't touch anything related that part. 
or `https://ceapex.pkgs.visualstudio.com/_packaging/LocTools/nuget/v3/index.json` might be the old one.  Do you know about it? 

```
OneLocBuildClient.exe Information: 0 : CredentialProvider.VSS: Couldn't get an authentication token for https://ceapex.pkgs.visualstudio.com/_packaging/LocTools/nuget/v3/index.json.
OneLocBuildClient.exe Information: 0 : CredentialProvider.VSS: Exit message: Auth tokens could not be retrieved. Remove the -NonInteractive flag and try again
OneLocBuildClient.exe Information: 0 : CredentialProvider.VSS: Exit code: Failure

```

pragnagopa (approved):

TsuyoshiUshio:
Hi @pragnagopa  according to the @soninaren , the CI build fails are known issue. We can ignore it.

PR #1082, opened 2021-08-17
Update DurableTask version to 2.*.* for v3.x

bachuv:
Updated Microsoft.Azure.WebJobs.Extensions.DurableTask version to `2.*.*`

### DIFF
--- a/Functions.Templates/Bindings/bindings.json
+++ b/Functions.Templates/Bindings/bindings.json
@@ -1477,7 +1477,7 @@
       "enabledInTryMode": false,
       "documentation": "$content=Documentation\\orchestrationClientIn.md",
       "extension": {
-        "id": "Microsoft.Azure.WebJobs.Extensions.DurableTask", "version": "2.1.1"
+        "id": "Microsoft.Azure.WebJobs.Extensions.DurableTask", "version": "2.*.*"
       },
       "settings": [
         {
--- a/Functions.Templates/Templates/DurableFunctionsActivity-CSharp-2.x/metadata.json
+++ b/Functions.Templates/Templates/DurableFunctionsActivity-CSharp-2.x/metadata.json
@@ -12,7 +12,7 @@
   "userPrompt": [],
   "extensions": [
     {
-      "id": "Microsoft.Azure.WebJobs.Extensions.DurableTask", "version": "2.1.1"
+      "id": "Microsoft.Azure.WebJobs.Extensions.DurableTask", "version": "2.*.*"
     }
   ]
 }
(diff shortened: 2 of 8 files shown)


Functions.Templates/Bindings/bindings.json
```diff
@@ -1477,7 +1477,7 @@
       "enabledInTryMode": false,
       "documentation": "$content=Documentation\\orchestrationClientIn.md",
       "extension": {
-        "id": "Microsoft.Azure.WebJobs.Extensions.DurableTask", "version": "2.1.1"
+        "id": "Microsoft.Azure.WebJobs.Extensions.DurableTask", "version": "2.*.*"
```
TsuyoshiUshio, line 1480: It looks 2.* is enough. :) https://docs.microsoft.com/en-us/nuget/concepts/package-versioning  https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-register#:~:text=Select%20a%20bundle%20version%20based%20on%20the%20extensions,Core%20Tools%20will%20automatically%20add%20it%20for%20you.
However, I'm ok if it works perfectly. What do you think? @soninaren

soninaren: For the change here: https://github.com/Azure/azure-functions-templates/blob/dev/Functions.Templates/Templates/DurableFunctionsEntityHttp-CSharp-2.x/.template.config/template.json#L38

The change may work and should not break the template as long as we don't break semver in durable releases. Assuming we have tested the template change in VS. I would have to reach out to the VS team to confirm this change is okay. One thing to note here is that NuGet package resolution has changed in the past from automatically resolving to the smallest version for dependencies to largest (or vice versa). Once this change is out it won't be in our control how this is resolved. The major down side is we won't be able to revert any unforeseen consequences of this change. We can control this on our end in core tools. Can't say the same for VS.

Also the change to bindings.json and metadata.json will break extension install experience in core tools. Here is output i got on a test run.
```
d:\azfun\user\tempFolders\2021818102>func extensions install -p "Microsoft.Azure.WebJobs.Extensions.DurableTask" -v "2.*.*"
  Determining projects to restore...
  Writing C:\Users\nasoni\AppData\Local\Temp\tmp1109.tmp
info : Adding PackageReference for package 'Microsoft.Azure.WebJobs.Extensions.DurableTask' into project 'd:\azfun\user\tempFolders\2021818102\extensions.csproj'.
error: '2.*.*' is not a valid version string.
```

liliankasem: @soninaren any updates on this?

soninaren:
Closing as invalid change

PR #1100, opened 2021-09-21
dropping preview1 suffix; updating to preview2 SDK

brettsam:


brettsam:
FYI on this @soninaren -- I'll go ahead and merge since I think this is pretty straightforward.

PR #1145, opened 2021-11-23
Update isolated worker template to default to V4 host

maartenba:
Because the project file has `net6.0` hard-coded into, and V3 + net6 don't go well together.


fabiocav:
@maartenba: @jongalloway has this change in his PR and we're about to merge that. Will close this in favor of that change, but please let us know if you have any concerns. 
Thanks for sending this!

PR #1191, opened 2022-04-12
Automatically generate Http port for Azure Functions project

vijayrkn:
This change is to ensure that more than one Azure functions project created from VS or the commandline can be run without any other modifications. 

With this change, every new Azure function project that gets created will be assigned an available port between 7000 & 7300. The default value for the port is 7071


vijayrkn:
This is the 2nd part of the change mentioned in https://github.com/Azure/azure-functions-templates/pull/1190

/cc @BillHiebert

phenning (approved):
Authoring looks perfect! :shipit:

soninaren (approved):

PR #1195, opened 2022-04-19
Bump template id so v3 and v4 can be used with `dotnet new`

maartenba:
Bump template id so v3 and v4 can be used with `dotnet new`. Right now the id is the same for both v3 and v4 templates.


maartenba:
Seems obsolete

PR #1200, opened 2022-05-06
Upgrade DF Extension from 2.5.1 to 2.7.0

davidmrdavid:
A long overdue change: We were updating branches v3 and v4 instead, but we just learned that dev branch is now the source of truth.


cgillum (commented):
I'm still seeing a 2.5.1 reference to our extension here: Functions.Templates/Bindings/bindings.json, line 1487.

davidmrdavid:
Just added it. Somehow my first CTRL+F missed it. Thanks!

@soninaren it would be great if we had these version references in a single place. Having them split across files makes it easy to miss one :)

PR #1202, opened 2022-05-09
add sql JS input and output binding templates

MaddyDev:
Added JavaScript templates for SQL Input and output Bindings

### DIFF
--- a/Build/PackageFiles/ExtensionBundle/ExtensionBundlePreviewTemplates-3.x.nuspec
+++ b/Build/PackageFiles/ExtensionBundle/ExtensionBundlePreviewTemplates-3.x.nuspec
@@ -49,7 +49,7 @@
     <file src="..\..\..\Functions.Templates\templates\DurableFunctionsHttpStart-PowerShell-2.x\**" target="templates\DurableFunctionsHttpStart-PowerShell" />
     <file src="..\..\..\Functions.Templates\templates\DurableFunctionsHttpStart-Python-2.x\**" target="Templates\DurableFunctionsHttpStart-Python" />
     <file src="..\..\..\Functions.Templates\templates\DurableFunctionsHttpStart-TypeScript\**" target="templates\DurableFunctionsHttpStart-TypeScript" />
-    
+
     <file src="..\..\..\Functions.Templates\templates\DurableFunctionsOrchestrator-CSharp-2.x\*.*" target="templates\DurableFunctionsOrchestrator-CSharp" exclude="**\*.cs"/>
     <file src="..\..\..\Functions.Templates\templates\DurableFunctionsOrchestrator-JavaScript\**" target="Templates\DurableFunctionsOrchestrator-JavaScript" />
     <file src="..\..\..\Functions.Templates\templates\DurableFunctionsOrchestrator-PowerShell-2.x\**" target="Templates\DurableFunctionsOrchestrator-PowerShell" />
@@ -69,7 +69,7 @@
     <file src="..\..\..\Functions.Templates\templates\EventHubTrigger-PowerShell\**" target="Templates\EventHubTrigger-PowerShell" />
     <file src="..\..\..\Functions.Templates\templates\EventHubTrigger-Python\**" target="templates\EventHubTrigger-Python" />
     <file src="..\..\..\Functions.Templates\templates\EventHubTrigger-TypeScript\**" target="templates\EventHubTrigger-TypeScript" />
-    
+
     <file src="..\..\..\Functions.Templates\templates\HttpTrigger-CSharp\*.*" target="templates\HttpTrigger-CSharp" exclude="**\*.cs" />
     <file src="..\..\..\Functions.Templates\templates\HttpTrigger-Custom\**" target="templates\HttpTrigger-Custom" />
     <file src="..\..\..\Functions.Templates\templates\HttpTrigger-JavaScript\**" target="templates\HttpTrigger-JavaScript" />
@@ -141,9 +141,9 @@
     <file src="..\..\..\Functions.Templates\templates\TimerTrigger-JavaScript\**" target="templates\TimerTrigger-JavaScript" />
     <file src="..\..\..\Functions.Templates\templates\TimerTrigger-PowerShell\**" target="templates\TimerTrigger-PowerShell" />
     <file src="..\..\..\Functions.Templates\templates\TimerTrigger-Python\**" target="templates\TimerTrigger-Python" />
-    <file src="..\..\..\Functions.Templates\templates\TimerTrigger-TypeScript\**" target="templates\TimerTrigger-TypeScript" /> 
+    <file src="..\..\..\Functions.Templates\templates\TimerTrigger-TypeScript\**" target="templates\TimerTrigger-TypeScript" />
 
-	  <file src="..\..\..\Functions.Templates\templates\SqlInputBinding-JavaScript\**" target="templates\SqlInputBinding-JavaScript" />
-    <file src="..\..\..\Functions.Templates\templates\SqlOutputBinding-JavaScript\**" target="templates\SqlOutputBinding-JavaScript" /> 	
+    <file src="..\..\..\Functions.Templates\templates\SqlInputBinding-JavaScript\**" target="templates\SqlInputBinding-JavaScript" />
+    <file src="..\..\..\Functions.Templates\templates\SqlOutputBinding-JavaScript\**" target="templates\SqlOutputBinding-JavaScript" />
   </files>
 </package>
--- a/Build/PackageFiles/ExtensionBundle/ExtensionBundlePreviewTemplates-3.x.nuspec
+++ b/Build/PackageFiles/ExtensionBundle/ExtensionBundlePreviewTemplates-3.x.nuspec
@@ -141,6 +141,9 @@
     <file src="..\..\..\Functions.Templates\templates\TimerTrigger-JavaScript\**" target="templates\TimerTrigger-JavaScript" />
     <file src="..\..\..\Functions.Templates\templates\TimerTrigger-PowerShell\**" target="templates\TimerTrigger-PowerShell" />
     <file src="..\..\..\Functions.Templates\templates\TimerTrigger-Python\**" target="templates\TimerTrigger-Python" />
-    <file src="..\..\..\Functions.Templates\templates\TimerTrigger-TypeScript\**" target="templates\TimerTrigger-TypeScript" />   
+    <file src="..\..\..\Functions.Templates\templates\TimerTrigger-TypeScript\**" target="templates\TimerTrigger-TypeScript" /> 
+
+	  <file src="..\..\..\Functions.Templates\templates\SqlInputBinding-JavaScript\**" target="templates\SqlInputBinding-JavaScript" />
+    <file src="..\..\..\Functions.Templates\templates\SqlOutputBinding-JavaScript\**" target="templates\SqlOutputBinding-JavaScript" /> 	
   </files>
 </package>
--- a/Build/PackageFiles/ExtensionBundle/ExtensionBundlePreviewTemplates-4.x.nuspec
+++ b/Build/PackageFiles/ExtensionBundle/ExtensionBundlePreviewTemplates-4.x.nuspec
@@ -49,7 +49,7 @@
     <file src="..\..\..\Functions.Templates\templates\DurableFunctionsHttpStart-PowerShell-2.x\**" target="templates\DurableFunctionsHttpStart-PowerShell" />
     <file src="..\..\..\Functions.Templates\templates\DurableFunctionsHttpStart-Python-2.x\**" target="Templates\DurableFunctionsHttpStart-Python" />
     <file src="..\..\..\Functions.Templates\templates\DurableFunctionsHttpStart-TypeScript\**" target="templates\DurableFunctionsHttpStart-TypeScript" />
-    
+
     <file src="..\..\..\Functions.Templates\templates\DurableFunctionsOrchestrator-CSharp-2.x\*.*" target="templates\DurableFunctionsOrchestrator-CSharp" exclude="**\*.cs"/>
     <file src="..\..\..\Functions.Templates\templates\DurableFunctionsOrchestrator-JavaScript\**" target="Templates\DurableFunctionsOrchestrator-JavaScript" />
     <file src="..\..\..\Functions.Templates\templates\DurableFunctionsOrchestrator-PowerShell-2.x\**" target="Templates\DurableFunctionsOrchestrator-PowerShell" />
@@ -69,7 +69,7 @@
     <file src="..\..\..\Functions.Templates\templates\EventHubTrigger-PowerShell\**" target="Templates\EventHubTrigger-PowerShell" />
     <file src="..\..\..\Functions.Templates\templates\EventHubTrigger-Python\**" target="templates\EventHubTrigger-Python" />
     <file src="..\..\..\Functions.Templates\templates\EventHubTrigger-TypeScript\**" target="templates\EventHubTrigger-TypeScript" />
-    
+
     <file src="..\..\..\Functions.Templates\templates\HttpTrigger-CSharp\*.*" target="templates\HttpTrigger-CSharp" exclude="**\*.cs" />
     <file src="..\..\..\Functions.Templates\templates\HttpTrigger-Custom\**" target="templates\HttpTrigger-Custom" />
     <file src="..\..\..\Functions.Templates\templates\HttpTrigger-JavaScript\**" target="templates\HttpTrigger-JavaScript" />
@@ -141,9 +141,9 @@
     <file src="..\..\..\Functions.Templates\templates\TimerTrigger-JavaScript\**" target="templates\TimerTrigger-JavaScript" />
     <file src="..\..\..\Functions.Templates\templates\TimerTrigger-PowerShell\**" target="templates\TimerTrigger-PowerShell" />
     <file src="..\..\..\Functions.Templates\templates\TimerTrigger-Python\**" target="templates\TimerTrigger-Python" />
-    <file src="..\..\..\Functions.Templates\templates\TimerTrigger-TypeScript\**" target="templates\TimerTrigger-TypeScript" /> 
+    <file src="..\..\..\Functions.Templates\templates\TimerTrigger-TypeScript\**" target="templates\TimerTrigger-TypeScript" />
 
-	  <file src="..\..\..\Functions.Templates\templates\SqlInputBinding-JavaScript\**" target="templates\SqlInputBinding-JavaScript" />
-    <file src="..\..\..\Functions.Templates\templates\SqlOutputBinding-JavaScript\**" target="templates\SqlOutputBinding-JavaScript" /> 	
+    <file src="..\..\..\Functions.Templates\templates\SqlInputBinding-JavaScript\**" target="templates\SqlInputBinding-JavaScript" />
+    <file src="..\..\..\Functions.Templates\templates\SqlOutputBinding-JavaScript\**" target="templates\SqlOutputBinding-JavaScript" />
   </files>
 </package>
--- a/Build/PackageFiles/ExtensionBundle/ExtensionBundlePreviewTemplates-4.x.nuspec
+++ b/Build/PackageFiles/ExtensionBundle/ExtensionBundlePreviewTemplates-4.x.nuspec
@@ -141,6 +141,9 @@
     <file src="..\..\..\Functions.Templates\templates\TimerTrigger-JavaScript\**" target="templates\TimerTrigger-JavaScript" />
     <file src="..\..\..\Functions.Templates\templates\TimerTrigger-PowerShell\**" target="templates\TimerTrigger-PowerShell" />
     <file src="..\..\..\Functions.Templates\templates\TimerTrigger-Python\**" target="templates\TimerTrigger-Python" />
-    <file src="..\..\..\Functions.Templates\templates\TimerTrigger-TypeScript\**" target="templates\TimerTrigger-TypeScript" />   
+    <file src="..\..\..\Functions.Templates\templates\TimerTrigger-TypeScript\**" target="templates\TimerTrigger-TypeScript" /> 
+
+	  <file src="..\..\..\Functions.Templates\templates\SqlInputBinding-JavaScript\**" target="templates\SqlInputBinding-JavaScript" />
+    <file src="..\..\..\Functions.Templates\templates\SqlOutputBinding-JavaScript\**" target="templates\SqlOutputBinding-JavaScript" /> 	
   </files>
 </package>
--- a/Functions.Templates/Templates/SqlInputBinding-JavaScript/function.json
+++ b/Functions.Templates/Templates/SqlInputBinding-JavaScript/function.json
@@ -1,10 +1,19 @@
 {
   "bindings": [
     {
+      "authLevel": "function",
+      "name": "req",
+      "type": "httpTrigger",
+      "direction": "in",
+      "methods": [
+        "get"
+      ]
+    },
+    {
       "type": "sql",
       "name": "results",
       "direction": "in",
-      "commandText": "select * from tableName",
+      "commandText": "select * from [dbo].[table1]",
       "commandType": "Text",
       "connectionStringSetting": "SqlConnectionString"
     }
--- a/Functions.Templates/Templates/SqlInputBinding-JavaScript/function.json
+++ b/Functions.Templates/Templates/SqlInputBinding-JavaScript/function.json
@@ -18,7 +18,7 @@
       "type": "sql",
       "name": "results",
       "direction": "in",
-      "commandText": "select * from [dbo].[table1]",
+      "commandText": "SELECT * FROM [dbo].[table1]",
       "commandType": "Text",
       "connectionStringSetting": "SqlConnectionString"
     }
--- a/Functions.Templates/Templates/SqlInputBinding-JavaScript/function.json
+++ b/Functions.Templates/Templates/SqlInputBinding-JavaScript/function.json
@@ -12,7 +12,7 @@
     {
       "type": "sql",
       "name": "results",
-      "direction": "in",
+      "direction": "out",
       "commandText": "select * from [dbo].[table1]",
       "commandType": "Text",
       "connectionStringSetting": "SqlConnectionString"
--- a/Functions.Templates/Templates/SqlInputBinding-JavaScript/function.json
+++ b/Functions.Templates/Templates/SqlInputBinding-JavaScript/function.json
@@ -10,9 +10,14 @@
       ]
     },
     {
+      "name": "$return",
+      "type": "http",
+      "direction": "out"
+    },
+    {
       "type": "sql",
       "name": "results",
-      "direction": "out",
+      "direction": "in",
       "commandText": "select * from [dbo].[table1]",
       "commandType": "Text",
       "connectionStringSetting": "SqlConnectionString"
--- a/Functions.Templates/Templates/SqlInputBinding-JavaScript/function.json
+++ b/Functions.Templates/Templates/SqlInputBinding-JavaScript/function.json
@@ -1,0 +1,12 @@
+{
+  "bindings": [
+    {
+      "type": "sql",
+      "name": "results",
+      "direction": "in",
+      "commandText": "select * from tableName",
+      "commandType": "Text",
+      "connectionStringSetting": "SqlConnectionString"
+    }
+  ]
+}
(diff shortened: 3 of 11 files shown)


Functions.Templates/Templates/SqlInputBinding-JavaScript/readme.md
```diff
@@ -0,0 +1,9 @@
+# SQL Input Binding - C<span>#</span>
```
chlafreniere, line 1: This should say JavaScript

Functions.Templates/Templates/SqlOutputBinding-JavaScript/index.js
```diff
@@ -0,0 +1,15 @@
+module.exports = async function (context, req) {
+    context.log('JavaScript HTTP trigger function processed a request.');
+
+    const testRow = {
+        "column1": column1Value,
```
dzsquared, line 5: not sure having the template trying to insert a null value is a good starting point.  can we set this var ahead of line 4 to ... something? the number 12 etc

MaddyDev: I can totally get rid of that and have the testRow directly assigned to the results?

Functions.Templates/Templates/SqlInputBinding-JavaScript/index.js
```diff
@@ -0,0 +1,5 @@
+module.exports = async function (context, results) {
+    if (!!results && results.length > 0) {
```
Charles-Gagnon, line 2: Why wouldn't we log if the results are 0? That's a perfectly valid response

(and can results ever be undefined? That seems like that shouldn't be a valid case. Are there other examples you're following that do this kind of check?)

Functions.Templates/Templates/SqlInputBinding-JavaScript/index.js
```diff
@@ -0,0 +1,5 @@
...
+        context.log('Query results: ', JSON.stringify(results));
```
Charles-Gagnon, line 3: We should do something besides just log - make sure you're returning the correct object so that the trigger shows the results as expected.

Functions.Templates/Templates/SqlOutputBinding-JavaScript/index.js
Charles-Gagnon: I don't see why we have our samples have these hardcoded values anyways. That isn't generally what people are going to be doing I would image - you should just create the row object from the param values passed in.

Functions.Templates/Templates/SqlOutputBinding-JavaScript/index.js
```diff
@@ -0,0 +1,15 @@
...
+        "column2": "column2Value"
+    };
+
+    context.bindings.results = req["body"]?.rows ?? JSON.stringify(testRow);
+
+    return {
+        status: 201,
+        body: context.bindings.results
```
Charles-Gagnon, line 14: `mimetype:"application/json"`

Functions.Templates/Templates/SqlInputBinding-JavaScript/index.js
MaddyDev: yeah am following the cosmosDB example.

Functions.Templates/Templates/SqlInputBinding-JavaScript/index.js
MaddyDev: I return the results in our samples, but again following the cosmosdb sample so didn't.

Functions.Templates/Templates/SqlInputBinding-JavaScript/index.js
Charles-Gagnon: I don't think we should assume they've done everything correctly (I've seen a lot of weird stuff in their examples) and it doesn't make sense to me to return undefined. Our code will always return *something* - even if it's just an empty list. 

And there's other samples like the blob trigger that don't do this so I think it should be fine to remove and help keep things cleaner. 

@chlafreniere @dzsquared ?

Functions.Templates/Templates/SqlInputBinding-JavaScript/index.js
Charles-Gagnon: Let's do what we think makes sense and not just copy them - given that our samples are pretty much all returning the items retrieved let's do that here too.

Functions.Templates/Templates/SqlOutputBinding-JavaScript/index.js
MaddyDev: I don't have this in our samples, return the mimetype as well?

Charles-Gagnon: Yes, we need to always be doing this. mimetype tells the browser how to render the body - so when we send json like this we need to tell it that to ensure things handle it correctly.

Functions.Templates/Templates/SqlInputBinding-JavaScript/function.json
```diff
@@ -0,0 +1,12 @@
+{
+  "bindings": [
```
dzsquared, line 2: am I reading correctly that the input binding template doesn't have a trigger?
can we be consistent with the C# templates, and add
```javascript
content.res = {
body: results,
// header for the mimetype
}
```

Functions.Templates/Templates/SqlInputBinding-JavaScript/index.js
dzsquared: results should not be undefined (bad things happened), but results.length could absolutely be 0.  instead of relying on logging, let's add an http trigger to this template and return the results.  (the logging can stay)

Functions.Templates/Templates/SqlInputBinding-JavaScript/function.json
MaddyDev: returned the results for both the bindings now.

Functions.Templates/Templates/SqlInputBinding-JavaScript/metadata.json
```diff
@@ -0,0 +1,18 @@
...
+  "defaultFunctionName": "SqlInputBinding",
+  "description": "$SqlInputBinding_description",
+  "name": "SQL Input Binding",
+  "language": "JavaScript",
+  "category": [
+    "$temp_category_core",
+    "$temp_category_dataProcessing"
+  ],
+  "categoryStyle": "sql",
+  "enabledInTryMode": false,
```
chlafreniere, line 11: do we know what TryMode is?

Functions.Templates/Templates/SqlInputBinding-JavaScript/function.json
```diff
@@ -0,0 +1,12 @@
...
+    {
+      "type": "sql",
+      "name": "results",
+      "direction": "in",
+      "commandText": "select * from tableName",
```
chlafreniere, line 7: In the readme, we talk about `[dbo].[table1]`. Should we be consistent here, @dzsquared, or does including the schema complicate things too much?

Functions.Templates/Templates/SqlInputBinding-JavaScript/metadata.json
MaddyDev: no, will have to check with Naren, but I have seen couple if examples and all of them have this value as false.

Functions.Templates/Templates/SqlInputBinding-JavaScript/function.json
Charles-Gagnon: My preference is always include the scheme (and quote both objects). It's just good practice and can help avoid potential mistakes so we should be encouraging that where possible.

Functions.Templates/Templates/SqlInputBinding-JavaScript/metadata.json
Charles-Gagnon: Probably https://www.tryfunctions.com/ng-min/try?trial=true (although that's never worked for me...)

Functions.Templates/Templates/SqlInputBinding-JavaScript/function.json
Charles-Gagnon: I'm still not seeing a trigger binding for this - we should add an HTTP trigger binding to match the C# templates

Functions.Templates/Templates/SqlOutputBinding-JavaScript/function.json
```diff
@@ -0,0 +1,12 @@
+{
+  "bindings": [
+    {
+      "type": "sql",
+      "name": "results",
+      "direction": "out",
+      "commandText": "[tableName]",
```
Charles-Gagnon, line 7: Schema here too

Functions.Templates/Templates/SqlOutputBinding-JavaScript/index.js
```diff
@@ -0,0 +1,13 @@
+module.exports = async function (context, req) {
+    context.log('JavaScript HTTP trigger function processed a request.');
+
+    var rows = req["body"]?.rows ?? [];
```
Charles-Gagnon, line 4: Given that most of our samples are for GET requests it seems better to also have the templates use that as well (i.e. use the URL parameters instead)

Functions.Templates/Templates/SqlOutputBinding-JavaScript/index.js
```diff
@@ -0,0 +1,13 @@
...
+    if (rows.length > 0) {
```
Charles-Gagnon, line 5: Shouldn't we be stringifying it regardless? The body should always be a string

Functions.Templates/Templates/SqlOutputBinding-JavaScript/index.js
MaddyDev: can we add multiple templates to the bindings, I thought we can only add one and post is what I used for all the JavaScript output binding examples.

Charles-Gagnon: I don't believe we can.

Why did you choose POST? Our C# output binding samples all use GET so it'd be better to be consistent with those.

MaddyDev: It doesn't make sense to use get for post requests (sending data out). It only make sense when we insert one row at a time, when doing multiple rows you have to use post anyway. I wasn't sure why C# samples were created that way but I can create a sample using params as an example and add it to be consistent. But I don't think that's the right way.

Charles-Gagnon: What do you mean "it doesn't make sense"? You can do both just fine. 

(and you can totally do multiple items with GET, you could just have a data blob that's a serialized array or something. Which at that point I agree using POST is much cleaner but it still *can* be done)

We should be consistent in our examples though - and GET is a lot easier to look at and understandso it makes sense for us to use that for the majority of our samples. But if we wanted to switch I don't really have a huge preference as long as we do the same for both. (and for the Python samples that I've made which mostly use GET)

Charles-Gagnon: @dzsquared @chlafreniere Any preference here?

dzsquared: I agree with @maddydev about using POST on the output binding and apologize for not catching that the C# output binding is GET only.
The default http trigger templates enable both GET and POST, with an example of checking for a request body before opting to a param value. That's a bit too much for our scenario, but it would be nice in the longer term to show output bindings as POST requests.

Charles-Gagnon: Are you ok with having our samples and templates use the different methods then? Because I am mostly saying here that we should be consistent - not that GET is better than POST in this scenario. Updating the templates/samples to support both seems like a good idea in the medium term though. 

If we're sticking with just POST then you should at least update function.json to indicate it only handles POST requests.

Functions.Templates/Templates/SqlOutputBinding-JavaScript/function.json
```diff
@@ -0,0 +1,22 @@
+{
+  "bindings": [
+    {
+      "authLevel": "function",
+      "name": "req",
+      "type": "httpTrigger",
+      "direction": "in",
+      "methods": [
+        "get",
```
Charles-Gagnon, line 9: Remove if we're not supporting GET in the sample

Functions.Templates/Templates/SqlOutputBinding-JavaScript/index.js
```diff
@@ -0,0 +1,13 @@
+module.exports = async function (context, req) {
```
Charles-Gagnon, line 1: `// Visit https://aka.ms/sqlbindingsoutput to learn how to use this output binding`

Functions.Templates/Templates/SqlOutputBinding-JavaScript/function.json
MaddyDev: we do support get, just that I haven't added a sample of output binding with get.

Charles-Gagnon: I'm saying that the sample doesn't support GET requests so we shouldn't have that included in the methods. Users are free to add it later on of course if they want, but we shouldn't be giving them a partially-completed function. 

(and then yes, later on when we fix this to support both GET and POST we can update it then)

Functions.Templates/Templates/SqlInputBinding-JavaScript/function.json
```diff
@@ -0,0 +1,21 @@
...
+      ]
+    },
+    {
+      "type": "sql",
+      "name": "results",
+      "direction": "in",
+      "commandText": "select * from [dbo].[table1]",
+      "commandType": "Text",
+      "connectionStringSetting": "SqlConnectionString"
+    }
```
Charles-Gagnon, line 19: This should also have the out response 

```
    {
      "type": "http",
      "direction": "out",
      "name": "res"
    }
```

Functions.Templates/Templates/SqlOutputBinding-JavaScript/function.json
```diff
@@ -0,0 +1,22 @@
...
+      ]
+    },
+    {
+      "type": "sql",
+      "name": "results",
+      "direction": "out",
+      "commandText": "[dbo].[table1]",
+      "commandType": "Text",
+      "connectionStringSetting": "SqlConnectionString"
+    }
```
Charles-Gagnon, line 20: This should have the out binding as well 

```
    {
      "type": "http",
      "direction": "out",
      "name": "res"
    }
```

Functions.Templates/Templates/SqlInputBinding-JavaScript/function.json
MaddyDev: for input binding ?

Charles-Gagnon: Yes. The response is for the HTTP trigger

Functions.Templates/Templates/SqlInputBinding-JavaScript/index.js
```diff
@@ -0,0 +1,12 @@
...
+    } else {
+        context.log("No results!");
+    }
+    return {
+        status: 201,
```
Charles-Gagnon, line 8: 201 is created, which is incorrect for an input binding. It should just be 200 "OK".

Functions.Templates/Templates/SqlInputBinding-JavaScript/index.js
```diff
@@ -0,0 +1,12 @@
...
+        mimetype:"application/json",
```
Charles-Gagnon, line 9: So I can't seem to find any actual documentation on this and it seems like none of the other templates are setting the type like this so let's just remove this since things seem to work just fine without it. 

(Python has a parameter for this but I'm not actually sure that JS does, or at least I'm not finding it)

Functions.Templates/Templates/SqlInputBinding-JavaScript/index.js
Charles-Gagnon: @MaddyDev You still have the results check - are you not going to remove it? This whole thing can just be simplified to 

```javascript
module.exports = async function (context, results) {
    context.log('Query results: ', JSON.stringify(results));
    return {
        status: 200,
        body: results
    };
}
```

Functions.Templates/Templates/SqlInputBinding-JavaScript/metadata.json
```diff
@@ -0,0 +1,18 @@
...
+  "userPrompt": [
+    "commandText",
+    "commandType",
+    "parameters",
```
Charles-Gagnon, line 15: I think we should remove the `parameters` and `commandType` prompts to simplify things - we don't want to confuse new people and there isn't a good way to tell users what those values mean or how to use them. 

@dzsquared Does that seem reasonable?

Functions.Templates/Templates/SqlOutputBinding-JavaScript/function.json
```diff
@@ -0,0 +1,26 @@
+{
+  "bindings": [
+    {
+      "authLevel": "function",
+      "name": "res",
+      "type": "http",
+      "direction": "out",
```
Charles-Gagnon, line 7: "in"

Functions.Templates/Templates/SqlOutputBinding-JavaScript/function.json
```diff
@@ -0,0 +1,26 @@
+{
+  "bindings": [
+    {
+      "authLevel": "function",
+      "name": "res",
```
Charles-Gagnon, line 5: `req`

Functions.Templates/Templates/SqlOutputBinding-JavaScript/index.js
```diff
@@ -0,0 +1,12 @@
...
+    context.log('JavaScript HTTP trigger function processed a request.');
```
Charles-Gagnon, line 2: I'd suggest simplifying this function to what I have for the TS samples : 

```javascript
    context.log('HTTP trigger with SQL output binding function processed a request.');

    // Set items array to context.bindings for insertion
    context.bindings.items = JSON.stringify(req.body);

    context.res = {
        status: 201,
        body: req.body
    };
```

Functions.Templates/Templates/SqlInputBinding-JavaScript/index.js
MaddyDev: the check is used just for logging, am returning the results either way. It can be simplified to above, @dzsquared want me to be consistent and do the above?

Charles-Gagnon: We want these samples to be useful and simple enough for users - having all these extra checks isn't necessary and just a distraction from the actual core part of the code that's doing useful stuff IMO

Functions.Templates/Templates/SqlInputBinding-JavaScript/metadata.json
Charles-Gagnon: Confirmed per the docs : https://github.com/Azure/azure-functions-templates#template-files-1

`    "enabledInTryMode": true,                  // Should this template be available in try mode: https://tryfunctions.com/ng-min/try?trial=true`

Charles-Gagnon: So we should leave this as false for now I think given that most users won't have a DB available that would be accessible by that. We can follow up with enabling it later if we find out a good way to do that.

Functions.Templates/Templates/SqlInputBinding-JavaScript/function.json
```diff
@@ -0,0 +1,26 @@
...
+    {
+      "name": "$return",
+      "type": "http",
+      "direction": "out"
+    },
+    {
+      "type": "sql",
+      "name": "results",
+      "direction": "in",
+      "commandText": "select * from [dbo].[table1]",
```
lucyzhang929, line 21: "SELECT * FROM [dbo].[table1]"

Functions.Templates/Templates/SqlInputBinding-JavaScript/metadata.json
```diff
@@ -0,0 +1,13 @@
+{
+  "defaultFunctionName": "SqlInputBindingJavaScript",
```
lucyzhang929, line 2: Just for consistency across all templates, do we want the function name to be just 'SqlInputBinding' or also include the language? @Charles-Gagnon @dzsquared

Functions.Templates/Templates/SqlInputBinding-JavaScript/readme.md
```diff
@@ -0,0 +1,9 @@
+# SQL Input Binding - JavaScript
+
+The `SQL Input Binding` makes it easy to retrieve data from a database, returning the output of the query or stored procedure to the function.
+
+## How it works
+
+For a `SQL Input Binding` to work, you can provide the query to retrive data from an existing object in the database. For instance, you can set the query to `Select * From [dbo].[table1]` to query `[dbo].[table1]` object in the commandText attribute.
```
lucyzhang929, line 7: I copied the readme from here to be consistent
https://github.com/Azure/azure-functions-templates/blob/d606cfcb697c3eff1a63eb52e65d1377120d6db9/Functions.Templates/Templates/SqlInputBinding-CSharp/readme.md?plain=1#L7

Functions.Templates/Templates/SqlInputBinding-JavaScript/metadata.json
Charles-Gagnon: Just SqlInputBinding - they (should) never have multiple bindings with different languages in a single project so no need to differentiate the names.

Build/PackageFiles/ExtensionBundle/ExtensionBundlePreviewTemplates-4.x.nuspec
```diff
@@ -141,6 +141,9 @@
     <file src="..\..\..\Functions.Templates\templates\TimerTrigger-JavaScript\**" target="templates\TimerTrigger-JavaScript" />
     <file src="..\..\..\Functions.Templates\templates\TimerTrigger-PowerShell\**" target="templates\TimerTrigger-PowerShell" />
     <file src="..\..\..\Functions.Templates\templates\TimerTrigger-Python\**" target="templates\TimerTrigger-Python" />
-    <file src="..\..\..\Functions.Templates\templates\TimerTrigger-TypeScript\**" target="templates\TimerTrigger-TypeScript" />   
+    <file src="..\..\..\Functions.Templates\templates\TimerTrigger-TypeScript\**" target="templates\TimerTrigger-TypeScript" /> 
+
+	  <file src="..\..\..\Functions.Templates\templates\SqlInputBinding-JavaScript\**" target="templates\SqlInputBinding-JavaScript" />
```
Charles-Gagnon, line 146: fix spacing

Build/PackageFiles/ExtensionBundle/ExtensionBundlePreviewTemplates-3.x.nuspec
```diff
@@ -141,6 +141,9 @@
     <file src="..\..\..\Functions.Templates\templates\TimerTrigger-JavaScript\**" target="templates\TimerTrigger-JavaScript" />
     <file src="..\..\..\Functions.Templates\templates\TimerTrigger-PowerShell\**" target="templates\TimerTrigger-PowerShell" />
     <file src="..\..\..\Functions.Templates\templates\TimerTrigger-Python\**" target="templates\TimerTrigger-Python" />
-    <file src="..\..\..\Functions.Templates\templates\TimerTrigger-TypeScript\**" target="templates\TimerTrigger-TypeScript" />   
+    <file src="..\..\..\Functions.Templates\templates\TimerTrigger-TypeScript\**" target="templates\TimerTrigger-TypeScript" /> 
+
+	  <file src="..\..\..\Functions.Templates\templates\SqlInputBinding-JavaScript\**" target="templates\SqlInputBinding-JavaScript" />
```
Charles-Gagnon, line 146: Fix spacing

Functions.Templates/Templates/SqlInputBinding-JavaScript/index.js
```diff
@@ -0,0 +1,20 @@
...
+ * See https://aka.ms/sqlbindingsinput for more information about using this binding
+ *
+ * These tasks should be completed prior to running :
+ *      1. Update "commandText" in function.json - this should be the query to execute to retrieve the values being returned
+ *      2. Add an app setting named "SqlConnectionString" containing the connection string
+ *          to use for the SQL connection
+ *      3. Change the bundle name in host.json to "Microsoft.Azure.Functions.ExtensionBundle.Preview" and the version to "[3.*, 4.0.0)"
+ * @param context The Azure Function runtime context
+ * @param req The HttpRequest that triggered this function
+ * @param items The array of objects returned by the SQL input binding
```
Charles-Gagnon, line 12: Either update `items` to be `results` or vice versa (I think `items` makes more sense)

Functions.Templates/Templates/SqlOutputBinding-JavaScript/function.json
Charles-Gagnon: ping, please remove GET as the sample doesn't currently support it. We can add it in later if we update the template to support both.

Functions.Templates/Templates/SqlOutputBinding-JavaScript/index.js
```diff
@@ -0,0 +1,25 @@
...
+ *      2. Add an app setting named "SqlConnectionString" containing the connection string
+ *          to use for the SQL connection
+ *      3. Change the bundle name in host.json to "Microsoft.Azure.Functions.ExtensionBundle.Preview" and the version to "[3.*, 4.0.0)"
+ * @param context The Azure Function runtime context
+ * @param req The HttpRequest that triggered this function
+ */
+module.exports = async function (context, req) {
+    context.log('HTTP trigger with SQL output binding function processed a request.');
+
+    // Set results array to context.bindings for insertion
```
Charles-Gagnon, line 16: `upsertion`

Build/PackageFiles/ExtensionBundle/ExtensionBundlePreviewTemplates-4.x.nuspec
MaddyDev: I have it formatted on my local copy and if I try to fix this by formatting, it's formatting the entire document, so I left it as is.

Charles-Gagnon: Just undo your changes and modify the line manually, you don't need to run format on the whole file. 

(you can even do this from Github if you want)

Charles-Gagnon: I also use a tool called [SourceTree](https://www.sourcetreeapp.com/) that lets you stage changes line-by-line so you can pick out what changes you want to push up.

Functions.Templates/Templates/SqlInputBinding-JavaScript/readme.md
Charles-Gagnon: @MaddyDev What @lucyzhang929 suggested would be good. What you have right now mentions the commandText attribute which isn't accurate for non-C# languages (as they don't use the attribute)

Functions.Templates/Templates/SqlOutputBinding-JavaScript/readme.md
```diff
@@ -0,0 +1,9 @@
+# SQL Output Binding - JavaScript
+
+The `SQL Output Binding` makes it easy to take an array of rows and upsert them into the user table (i.e. If a row doesn't already exist, it is added. If it does, it is updated).
+
+## How it works
+
+For a `SQL Output Binding` to work, you provide the existing table in the database to upsert rows into. For instance, you can set the table to `[dbo].[table1]` in the commandText attribute to upsert into this table.
```
Charles-Gagnon, line 7:
```
For a `SQL Output Binding` to work, you provide the existing table in the database to upsert rows into. For instance, you can set the table to `[dbo].[table1]` in the `commandText` value in your `function.json` to upsert into this table.
```

Non-C# languages don't have the attribute so this description isn't accurate currently.

Functions.Templates/Templates/SqlInputBinding-JavaScript/readme.md
Charles-Gagnon: And I'd actually suggest using what I just pushed up for TS to be clearer 

```
For a `SQL Input Binding` to work, you can provide the query to retrieve data from an existing object in the database. For instance, you can set the query to `SELECT * FROM [dbo].[table1]` in the `commandText` value in your `function.json` to return all the values from the `[dbo].[table1]` table in the database.
```

MaddyDev: Updated it.

Charles-Gagnon (approved):

PR #1251, opened 2022-09-07
Reverting Event Hub Python template to use many cardinality

### LINKED ISSUE #1250: EventHub trigger for Python has wrong cardinality
When creating an event hub trigger using Python, there is a cardinality mismatch.

The `function.json` has the following:

```json
{
    "scriptFile": "__init__.py",
    "bindings": [
      {
        "type": "eventHubTrigger",
        "name": "event",
        "direction": "in",
        "eventHubName": "samples-workitems",
        "connection": "",
        "cardinality": "many",
        "consumerGroup": "$Default"
      }
    ]
}
```

Notice `"cardinality": "many"`/

In the code in `__init__.py`, the expected cardinality is 1:

```python
def main(event: func.EventHubEvent):
```

This needs to be fixed - the previous version had the correct cardinality:

```python
def main(events: List[func.EventHubEvent]):
```

As this stands, creating an event hub trigger produces a broken trigger.

jimbobbennett:
The event hub Python trigger was recently updated to have single cardinality, but the `cardinality` field itself was not updated, meaning creating a new event hub python trigger would lead to a functions app that would crash.

This PR reverts the single cardinality back to many, keeping this trigger in-line with the other event hub triggers. Fixes
#1250

### DIFF
--- a/Functions.Templates/Templates/EventHubTrigger-Python/__init__.py
+++ b/Functions.Templates/Templates/EventHubTrigger-Python/__init__.py
@@ -1,8 +1,10 @@
+from typing import List
 import logging
 
 import azure.functions as func
 
 
-def main(event: func.EventHubEvent):
-    logging.info('Python EventHub trigger processed an event: %s',
-                 event.get_body().decode('utf-8'))
+def main(events: List[func.EventHubEvent]):
+     for event in events:
+         logging.info('Python EventHub trigger processed an event: %s',
+                         event.get_body().decode('utf-8'))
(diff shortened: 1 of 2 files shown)


jimbobbennett:
Tagging @soninaren as PRs only seem to build when they raise them. @soninaren - if there's no way to fix builds so anyone other than you can raise PRs, can you re-raise this yourself?

jimbobbennett:
These were tested by building locally and testing with Azure Functions Core tools.

PR #1271, opened 2022-10-14
Default blob extension to track2

hossam-nasr:


soninaren:
@hossam-nasr were you able to test the templates using the instructions here
https://github.com/Azure/azure-functions-templates#testing-dotnet-templates

soninaren (approved):

PR #1281, opened 2022-11-03
Add missing C# DF Entity templates to the v3 and v4 ItemTemplate files

davidmrdavid:
We recently learned that none of our Entity templates for C# are showing up in VSCode, in any of the template categories.
I **suspect** this is because those templates were not listed in the `ItemTemplates` files in this repo, which would have been an oversight.

@soninaren: can you please confirm that this theory makes sense?
Additionally, I noticed that these files are under a directory called "Dotnet_precompiled". What does the "precompiled" mean in this context?


soninaren (approved):

davidmrdavid:
Got PM's ACK. Merging!

PR #1290, opened 2022-11-28
Update DF version to 2.9.0

nytiannn:
As title, routine PR.


davidmrdavid (approved):
LGTM!

davidmrdavid:
@soninaren could you please review and merge if appropiate?

bachuv:
Just want to call out that this PR also updates `bindings-bundle-v4.json`. We didn't update this previously, but I believe this is a new file and should be updated as well. @soninaren can you please confirm that we updated the right files? Thanks!

kashimiz (approved):

PR #1296, opened 2022-12-08
Default EventHub extension to track2

hossam-nasr:


hossam-nasr:
Closed in favor of #1317

PR #1301, opened 2022-12-12
Bump Event Grid extension version

JoshLove-msft:
CloudEvent handshake support was added in 3.2.0. Updating the version here will allows CloudEvent schema event grid subscriptions to be used for new EG functions created in the portal.


JoshLove-msft:
Just pinging on this. Any updates on the failing CI?

soninaren (approved):

PR #1320, opened 2023-01-25
Add durable isolated orchestration item template

jviau:
Adds item template for durable functions orchestration in dotnet isolated. The template is a direct copy of the existing template for in-proc, with changes to make it compatible with isolated (using and type changes).


Functions.Templates/Templates/DurableFunctionsOrchestration-CSharp-Isolated/DurableFunctionsOrchestrationCSharp.cs
```diff
@@ -0,0 +1,54 @@
...
+
+namespace Company.Function
+{
+    public static class DurableFunctionsOrchestrationCSharp
+    {
+        [Function(nameof(DurableFunctionsOrchestrationCSharp))]
+        public static async Task<List<string>> RunOrchestrator(
+            [OrchestrationTrigger] TaskOrchestrationContext context)
+        {
+            ILogger logger = context.CreateReplaySafeLogger(nameof(DurableFunctionsOrchestrationCSharp));
```
cgillum, line 14: I like that you added the logger as part of this template!

Functions.Templates/Templates/DurableFunctionsOrchestration-CSharp-Isolated/DurableFunctionsOrchestrationCSharp.cs
```diff
@@ -0,0 +1,54 @@
+using Microsoft.Azure.Functions.Worker;
+using Microsoft.Azure.Functions.Worker.Http;
+using Microsoft.DurableTask;
+using Microsoft.Extensions.Logging;
+
+namespace Company.Function
+{
+    public static class DurableFunctionsOrchestrationCSharp
```
cgillum, line 8: This might be a dumb question, but are there any template parameters we need to consider and use for the class name? Or is it okay that we hardcode it and let developers rename it after its created?

cgillum (commented):
I'm not really familiar with how item templates work so I can't really comment on the JSON content, but everything else pretty much looks good to me. I did have one question below, though:

Functions.Templates/Templates/DurableFunctionsOrchestration-CSharp-Isolated/DurableFunctionsOrchestrationCSharp.cs
jviau: That is a great question haha. I think we could definitely improve this template to parameterize the name. However, given the time constraints I would prefer to leave that for a later time.

cgillum (approved):
Signing off with the caveat that I'm not at all an expert on these templates. We should probably get an ACK from @soninaren before merging.

Functions.Templates/Templates/DurableFunctionsOrchestration-CSharp-Isolated/DurableFunctionsOrchestrationCSharp.cs
soninaren: The string `DurableFunctionsOrchestrationCSharp` is actually value of the `source` property. VS actually prompts for the functionName and replaces the source property with user input.

Functions.Templates/Templates/DurableFunctionsOrchestration-CSharp-Isolated/.template.config/template.json
```diff
@@ -0,0 +1,73 @@
...
+  "defaultName": "DurableFunctionsOrchestrationCSharp",
+  "postActions": [
+    {
+      "Description": "Adding reference to Microsoft.Azure.Functions.Worker.Extensions.DurableTask NuGet package.",
+      "ActionId": "B17581D1-C5C9-4489-8F0A-004BE667B814",
+      "ContinueOnError": "true",
+      "ManualInstructions": [],
+      "args": {
+        "referenceType": "package",
+        "reference": "Microsoft.Azure.Functions.Worker.Extensions.DurableTask", "version": "1.0.0",
```
soninaren, line 41: I see this package is not yet published. Do we have planned date for publishing this?

jviau: Yes - either later today or tomorrow.

soninaren (approved):

Functions.Templates/Templates/DurableFunctionsOrchestration-CSharp-Isolated/.template.config/template.json
jviau: They are published now https://www.nuget.org/packages/Microsoft.Azure.Functions.Worker.Extensions.DurableTask/1.0.0

PR #1327, opened 2023-02-10
Update Microsoft.Azure.WebJobs.Extensions.DurableTask version to 2.9.1

bachuv:
Updated Microsoft.Azure.WebJobs.Extensions.DurableTask version from 2.9.0 to 2.9.1


davidmrdavid (approved):
LGTM. Also @soninaren: is there a way to make the templates always get the latest extension in range 2.x?

soninaren (approved):

PR #1336, opened 2023-02-22
Feature/kusto templates

ag-ramachandran:


ag-ramachandran:
Built and tried for VS as provided in the steps.

InputBinding
OutputBinding

soninaren (approved):

PR #1339, opened 2023-02-26
Fixing multiple issues in the templates.

khkh-ms:
Fixing V2 templates.

### DIFF
--- a/Functions.Templates/Templates-v2/HttpTrigger-Python/function_body.py
+++ b/Functions.Templates/Templates-v2/HttpTrigger-Python/function_body.py
@@ -1,5 +1,5 @@
 @app.function_name(name="$(FUNCTION_NAME_INPUT)")
-@app.route(route="hello")
+@app.route(route="$(ROUTE_VALUE)")
 def $(FUNCTION_NAME_INPUT)(req: func.HttpRequest) -> func.HttpResponse:
      logging.info('Python HTTP trigger function processed a request.')
 
--- a/Functions.Templates/Templates-v2/TimerTrigger-Python/template.json
+++ b/Functions.Templates/Templates-v2/TimerTrigger-Python/template.json
@@ -24,15 +24,13 @@
         },
         {
             "name": "Append Timer Function to File",
-            "input": [
-                {
+            "input": {
                     "userCommand": "appendToFile",
                     "assignTo": "$(SELECTED_FILENAME)"
-                }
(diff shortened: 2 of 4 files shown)


Functions.Templates/Templates-v2/HttpTrigger-Python/function_body.py
```diff
@@ -1,5 +1,5 @@
 @app.function_name(name="$(FUNCTION_NAME_INPUT)")
-@app.route(route="hello")
+@app.route(route="$(ROUTE_VALUE)")
```
kashimiz, line 2: Do you want the `route` parameter to also take user input in the `Create New Project with HttpTrigger function` job?

Functions.Templates/Templates-v2/TimerTrigger-Python/template.json
```diff
@@ -89,7 +87,7 @@
             "name": "appendFileContent_FunctionApp",
             "type": "AppendToFile",
             "createIfNotExists" : false,
-            "source": "$(FUNCTION_APP_TRANSFORMED)",
+            "source": "$(TIMER_FUNCTION_BODY)",
```
kashimiz, line 90: Thanks for catching this. `$(AppBody)` should also be changed to `$(TIMER_FUNCTION_BODY)` in `function_app.py`.

kashimiz (approved):
Looks good; just caught one additional fix needed.

Also, we should probably change the `author` of the templates to whoever will be responsible for their upkeep before they become 'official.' Would this be @gavin-aguiar ?

Functions.Templates/Templates-v2/TimerTrigger-Python/template.json
khkh-ms: So far, I have only tried it "appendToFile" actions therefore didn't come across the mentioned issue.

Functions.Templates/Templates-v2/HttpTrigger-Python/function_body.py
soninaren: I think it might be a good idea to use function name `FUNCTION_NAME_INPUT` for route, as it would be unique and would correlate with function as well. Also it would match the current templating experience as well as would be less confusing for a completely new user as they may not be familiar with functions.

soninaren (changes requested):

PR #1349, opened 2023-03-17
V2 Templates Schema: Action Docs

### LINKED ISSUE #1330: V2 Schema: Create reference documentation for actions.
Part of #1311

In the style of [.NET API docs.](https://learn.microsoft.com/en-us/dotnet/api/system.io.file.open?view=net-7.0)

kashimiz:
Resolves #1330

* A couple of example template fixes.
* Moved schemas to their own folder under Docs.
* Docs for all currently defined actions.
* Section on conditionals.

Some questions:
- should we allow placeholders in a conditional's `name` or `values` properties?
- `UserInput` needs an example; would you mind adding this?


Functions.Templates/Templates-v2/TimerTrigger-Python/template.json
```diff
@@ -57,7 +57,7 @@
                     "type": "AppendToFile",
```
gavin-aguiar, line 57: Actions within a job should just be a list of strings right? This has a json. Shouldn't this json be in the actions section below?

Docs/Actions/ReplaceTokensInText.md
```diff
@@ -0,0 +1,163 @@
...
+| Environment  | Support |
+| ------------ | ------- |
+| Azure Portal | :white_check_mark: |
+| Core Tools   | :white_check_mark: |
+| VS Code      | :white_check_mark: |
+
+## Parameters
+
+**`type`**  
+**`ReplaceTokensInText`**
```
gavin-aguiar, line 18: If writeToFile and appendToFile have a property called replaceToken then why  do we need this action?

Functions.Templates/Templates-v2/HttpTrigger-Python/template.json
```diff
@@ -8,10 +8,10 @@
         {
             "name": "Create New Project with HttpTrigger function",
             "type": "CreateNewApp",
-            "input" : [
+            "inputs" : [
                 {
                     "assignTo": "$(APP_FILENAME)",
```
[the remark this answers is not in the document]
gavin-aguiar: Also if the value of this variable is function_app.py, can it be saved as a global variable?
We would need to access this filename while adding blueprints. When a blueprint is added, we need to register the blueprint in the APP_FILENAME file. Since in the adding blueprint job, there is not input to read the APP_FILENAME, we need to way to preserve the state if this variable so that it can be used later.

Docs/Actions/ReplaceTokensInText.md
kashimiz: This is covered in the Remarks section of this document; this action is for future template scenarios that may require a token replacement step that isn't paired with a file write.

Functions.Templates/Templates-v2/TimerTrigger-Python/template.json
kashimiz: This is an example of an anonymous action. There should be a section explaining when and how you would define one in the internal walkthrough doc (which I should pull over here) but it's an alternative way of defining actions specific to a single job.

Functions.Templates/Templates-v2/HttpTrigger-Python/template.json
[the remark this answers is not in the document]
kashimiz: Discussed various strategies to approach this offline.

soninaren (approved):

PR #1397, opened 2023-06-22
Add sql templates to non preview extension bundle nuspec

### LINKED ISSUE #875: Missing Sql input/output templates when creating through func cli
Missed adding the sql template updates to non preview extension bundle nuspec so the sql templates are still not showing when trying to use func cli.

MaddyDev:
Fixes https://github.com/Azure/azure-functions-sql-extension/issues/875


VasuBhog (approved):
This is needed to help show verified templates for both cli and vscode-azurefunctions.

lucyzhang929 (approved):

PR #1399, opened 2023-06-22
Update blob trigger template

aishwaryabh:


Functions.Templates/Templates/BlobTrigger-CSharp-Isolated/.template.config/template.json
```diff
@@ -51,7 +51,7 @@
             "ManualInstructions": [],
             "args": {
                 "referenceType": "package",
-            "reference": "Microsoft.Azure.Functions.Worker.Extensions.Storage", "version": "4.0.4",
+            "reference": "Microsoft.Azure.Functions.Worker.Extensions.Storage", "version": "5.1.1-preview2",
```
kshyju, line 54: Do we usually include _preview_ version of packages in our templates? I was under the impression that we do not.

aishwaryabh: We wanted to update the template to reflect the sdk type binding changes and that's in preview today.

aishwaryabh:
Replacing this PR with another one due to weird dependency issues

Functions.Templates/Templates/BlobTrigger-CSharp-Isolated/.template.config/template.json
liliankasem: We don't ship templates with preview versions, this change will ultimately be for the GA package. So we should test the templates with the preview package but update to the stable (GA) package before merging

PR #1412, opened 2023-07-06
Add Node.js model v4, schema v2 templates

ejizba:
These should be identical to the templates we used for preview, just using the new v2 schema. I modeled these after the python new model templates, the schema [here](https://github.com/Azure/azure-functions-templates/tree/dev/Docs), and the docs [here](https://eng.ms/docs/cloud-ai-platform/devdiv/serverless-paas-balam/serverless-paas-benbyrd/app-service-web-apps/app-service-team-documents/functionteamdocs/design/templateexperience/templateschemav2). I was not able to get a full end-to-end test with these in any of our tooling because the tooling just doesn't seem ready yet. I did some _super rough_ testing in VS Code (with @nturinski's help), and it seemed to be on the right track. cc @khkh-ms for core tools and @shimedh for portal.

Here's the full templates-v2.json if any of the tooling folks want to do their own testing:
https://gist.github.com/ejizba/d41313bd36818ded13452b8f70ce406b

### DIFF
--- a/Functions.Templates/Templates-v2/BlobTrigger-JavaScript/index.js
+++ b/Functions.Templates/Templates-v2/BlobTrigger-JavaScript/index.js
@@ -1,0 +1,9 @@
+const { app } = require('@azure/functions');
+
+app.storageBlob('$(FUNCTION_NAME_INPUT)', {
+    path: '$(PATH_INPUT)',
+    connection: '$(CONNECTION_INPUT)',
+    handler: (blob, context) => {
+        context.log(`Storage blob function processed blob "${context.triggerMetadata.name}" with size ${blob.length} bytes`);
+    }
+});
--- a/Functions.Templates/Templates-v2/BlobTrigger-JavaScript/template.json
+++ b/Functions.Templates/Templates-v2/BlobTrigger-JavaScript/template.json
@@ -1,0 +1,47 @@
+{
+    "author": "Eric Jizba",
+    "name": "Azure Blob Storage trigger",
+    "description": "$BlobTrigger_description",
+    "programmingModel": "v4",
+    "language": "JavaScript",
+    "jobs": [
+        {
+            "name": "Create New Function",
+            "type": "CreateNewApp",
+            "inputs": [
+                {
+                    "assignTo": "$(FUNCTION_NAME_INPUT)",
+                    "paramId": "functionName",
+                    "defaultValue": "storageBlobTrigger"
+                },
+                {
+                    "assignTo": "$(CONNECTION_INPUT)",
+                    "paramId": "connection"
+                },
+                {
+                    "assignTo": "$(PATH_INPUT)",
+                    "paramId": "path"
+                }
+            ],
+            "actions": [
+                "readFile",
+                "writeFile"
+            ]
+        }
+    ],
+    "actions": [
+        {
+            "name": "readFile",
+            "type": "GetTemplateFileContent",
+            "assignTo": "$(INDEX_FILE)",
+            "filePath": "index.js"
+        },
+        {
+            "name": "writeFile",
+            "type": "WriteToFile",
+            "source": "$(INDEX_FILE)",
+            "filePath": "src/functions/$(FUNCTION_NAME_INPUT).js",
+            "replaceTokens": true
+        }
+    ]
+}
--- a/Functions.Templates/Templates-v2/BlobTrigger-JavaScript/template.json
+++ b/Functions.Templates/Templates-v2/BlobTrigger-JavaScript/template.json
@@ -11,27 +11,27 @@
             "inputs": [
                 {
                     "assignTo": "$(FUNCTION_NAME_INPUT)",
-                    "paramId": "functionName",
+                    "paramId": "trigger-functionName",
                     "defaultValue": "storageBlobTrigger"
                 },
                 {
                     "assignTo": "$(CONNECTION_INPUT)",
-                    "paramId": "connection"
+                    "paramId": "blobTrigger-connection"
                 },
                 {
                     "assignTo": "$(PATH_INPUT)",
-                    "paramId": "path"
+                    "paramId": "blobTrigger-path"
                 }
             ],
             "actions": [
-                "readFile",
+                "getTemplateFile",
                 "writeFile"
             ]
         }
     ],
     "actions": [
         {
-            "name": "readFile",
+            "name": "getTemplateFile",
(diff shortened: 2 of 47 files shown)


Functions.Templates/Templates-v2/BlobTrigger-JavaScript/index.js
```diff
@@ -0,0 +1,9 @@
+const { app } = require('@azure/functions');
```
hossam-nasr, line 1: Should we also add these `.md` files, as in the docs?

Functions.Templates/Templates-v2/BlobTrigger-JavaScript/template.json
```diff
@@ -0,0 +1,47 @@
+{
+    "author": "Eric Jizba",
+    "name": "Azure Blob Storage trigger",
+    "description": "$BlobTrigger_description",
+    "programmingModel": "v4",
+    "language": "JavaScript",
+    "jobs": [
+        {
+            "name": "Create New Function",
+            "type": "CreateNewFile",
```
hossam-nasr, line 10: Looking at the docs, there doesn't seem to be a supported `CreateNewFile` option, only `CreateNewApp` and `AppendToFile`. Am I missing something?

Also, are we not supporting the appending scenario?

Functions.Templates/Templates-v2/BlobTrigger-JavaScript/template.json
```diff
@@ -0,0 +1,47 @@
...
+    "programmingModel": "v4",
+    "language": "JavaScript",
+    "jobs": [
+        {
+            "name": "Create New Function",
+            "type": "CreateNewFile",
+            "inputs": [
+                {
+                    "assignTo": "$(FUNCTION_NAME_INPUT)",
+                    "paramId": "functionName",
```
hossam-nasr, line 14: I don't see a `functionName` in [`userPrompts.json`](https://github.com/Azure/azure-functions-templates/blob/dev/Functions.Templates/Bindings/userPrompts.json). Perhaps you meant `trigger-functionName`?

Functions.Templates/Templates-v2/BlobTrigger-JavaScript/template.json
```diff
@@ -0,0 +1,47 @@
...
+                    "defaultValue": "storageBlobTrigger"
+                },
+                {
+                    "assignTo": "$(CONNECTION_INPUT)",
+                    "paramId": "connection"
```
hossam-nasr, line 19: Should be `blobTrigger-connection`

Functions.Templates/Templates-v2/BlobTrigger-JavaScript/template.json
```diff
@@ -0,0 +1,47 @@
...
+                },
+                {
+                    "assignTo": "$(PATH_INPUT)",
+                    "paramId": "path"
```
hossam-nasr, line 23: Should be `blobTrigger-path`

Functions.Templates/Templates-v2/BlobTrigger-JavaScript/template.json
```diff
@@ -0,0 +1,47 @@
...
+            ],
+            "actions": [
+                "readFile",
+                "writeFile"
+            ]
+        }
+    ],
+    "actions": [
+        {
+            "name": "readFile",
```
hossam-nasr, line 34: Nit: I would recommend a different name for this. I think "readFile" implied to me that we were reading the _customer's_ file and then writing to it, or something like that. I'd suggest something like "readTemplate" or "getTemplateContents" or something like that

Functions.Templates/Templates-v2/BlobTrigger-JavaScript/template.json
```diff
@@ -0,0 +1,47 @@
...
+                    "paramId": "connection"
+                },
+                {
+                    "assignTo": "$(PATH_INPUT)",
+                    "paramId": "path"
+                }
+            ],
+            "actions": [
+                "readFile",
+                "writeFile"
```
hossam-nasr, line 28: Do we need to add a `ReplaceTokensInText` step?

hossam-nasr (commented):
Left a few comments. I commented just on the blob trigger template since it was the first one, but my comment apply to all templates as well

liliankasem:
@ejizba it's been a month, do you think the core tools ready so you can test this E2E?

ejizba:
@liliankasem thanks for checking. The portal is the priority for this particular PR, because core tools and VS Code already have a workaround to use old-schema templates with the new model. During portal testing, we ran into quite a few blockers and have been discussing them offline with the portal team.

I have not tried core tools, but until we figure out the portal issues this PR will be blocked anyways.

Functions.Templates/Templates-v2/BlobTrigger-JavaScript/index.js
ejizba: Are you talking about the `.md` files that the Python team added? Let me know if you feel strongly, but I'm not a fan. First, they feel like an annoying pop-up when used in VS Code and AFAIK there's no way to do "Don't show again". Second, I think they're liable to get out-of-date and I would prefer if the official microsoft docs are the single source of truth. Last, my hope is that our template experience is good enough that users shouldn't need a page of docs just to try stuff out

Functions.Templates/Templates-v2/BlobTrigger-JavaScript/template.json
ejizba: Updated, thanks! I was using the old schema ids

Functions.Templates/Templates-v2/BlobTrigger-JavaScript/template.json
ejizba: updated

liliankasem:
> @liliankasem thanks for checking. The portal is the priority for this particular PR, because core tools and VS Code already have a workaround to use old-schema templates with the new model. During portal testing, we ran into quite a few blockers and have been discussing them offline with the portal team.
> 
> I have not tried core tools, but until we figure out the portal issues this PR will be blocked anyways.

Thanks for the update! Fill flag this as blocked for now

ejizba:
We decided not to support the portal for the new model, so in that sense this is no longer blocked. We still want to update core tools/vs code to use these new templates, but as mentioned they already have a workaround so this was de-prioritized. I can close this PR until we're actively working on it

PR #1421, opened 2023-07-12
Adding the net8 changes to CSharp-Isolated template.config

kshyju:
Adding the net8 changes to CSharp-Isolated template.config.

### DIFF
--- a/Functions.Templates/ProjectTemplate_v4.x/CSharp-Isolated/.template.config/template.json
+++ b/Functions.Templates/ProjectTemplate_v4.x/CSharp-Isolated/.template.config/template.json
@@ -32,6 +32,10 @@
                 "description": "Target .NET 7"
                 },
                 {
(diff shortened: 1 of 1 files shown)


Functions.Templates/ProjectTemplate_v4.x/CSharp-Isolated/.template.config/template.json
```diff
@@ -32,6 +32,10 @@
                 "description": "Target .NET 7"
                 },
                 {
+                "choice": "net8.0",
```
kshyju, line 35: I think this is the only file I need to update for .net 8 support in this repo. Once merged and we release a nuget package which include this change, will use that in the cli feed to continue my testing.

satvu (approved):

liliankasem (approved):

PR #1429, opened 2023-07-24
Adding EnableLiveMetricsFilters in host.json

surgupta-msft:
Issue - https://github.com/Azure/azure-functions-host/issues/9408

Add new setting `EnableLiveMetricsFilters` in host.json

### DIFF
--- a/Functions.Templates/ProjectTemplate_v4.x/CSharp/host.json
+++ b/Functions.Templates/ProjectTemplate_v4.x/CSharp/host.json
@@ -5,7 +5,8 @@
             "samplingSettings": {
                 "isEnabled": true,
                 "excludedTypes": "Request"
-            }
+            },
+            "enableLiveMetricsFilters": false
         }
     }
 }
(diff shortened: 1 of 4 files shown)


soninaren (approved):

Functions.Templates/ProjectTemplate_v4.x/CSharp/host.json
```diff
@@ -5,7 +5,8 @@
             "samplingSettings": {
                 "isEnabled": true,
                 "excludedTypes": "Request"
-            }
+            },
+            "enableLiveMetricsFilters": true
```
surgupta-msft, line 9: @brettsam wanted to confirm that we need this setting only for in-proc app and also the value should be true (i.e. by default we want live metrics to be _filtered_)

brettsam: we want it for all apps as the logs are really coming from the extensions (which all app types use) -- and yes, we want the filters to apply to live metrics

surgupta-msft: Added to other templates.

liliankasem (approved):

PR #1430, opened 2023-07-25
changing AzureWebJobsStorage data type to valid type

mattchenderson:
VS issue processing existing type, which we don't think is depended upon anywhere, but we should check, hence leaving this draft for now.


mattchenderson:
Both VS and VS Code have cleared this, so marking ready for review.

liliankasem (approved):

kshyju (approved):

PR #1438, opened 2023-08-15
Add Event Grid Cloud Event C# and F# templates

hossam-nasr:
Adds C# in-proc, F# in-proc, and C# script templates for Event Grid triggers using "Cloud Event" binding and schema. Resolves [this WI](https://dev.azure.com/FunctionsExtensions/Functions%20Extensions/_boards/board/t/Functions%20Extensions%20Team/Issues/?workitem=106), requested by the Event Grid team


soninaren (approved):
Looks good to me. I see that you have function.json and metadata.json files in the templates. You don't need them if the intent is to only add precompiled templates. Otherwise we should remove them

PR #1464, opened 2023-10-05
Renamed schedule to timer_trigger for py functions

gavin-aguiar:
Updated timer trigger functions for python v2.
timer_trigger is the new alias for schedule for v2 timer trigger templates

the runOnStartup property for timer triggers determines whether the function should run immediately when the function app starts, rather than waiting for the scheduled time. By default, this property is set to false. There are several reasons for this default setting:

- Timer functions are typically used for scheduled tasks based on a specific timetable (e.g., every hour, daily, etc.). Setting runOnStartup to false ensures that the function runs according to its defined schedule, maintaining the predictability of executions.

- If runOnStartup were true by default, every time the function app restarts (which can happen for various reasons like deployments, platform scaling, etc.), the timer-triggered function would execute. 

- Azure Functions can be dynamically scaled, and new instances can be started based on demand. With runOnStartup set to false, new instances won't immediately execute timer-triggered functions

### DIFF
--- a/Docs/Actions/ReplaceTokensInText.md
+++ b/Docs/Actions/ReplaceTokensInText.md
@@ -142,7 +142,7 @@ $(TIMER_FUNCTION_BODY)
 
 ```python
 @app.function_name(name="$(FUNCTION_NAME_INPUT)")
-@app.schedule(schedule="$(SCHEDULE_INPUT)", arg_name="myTimer", run_on_startup=True,
+@app.timer_trigger(schedule="$(SCHEDULE_INPUT)", arg_name="myTimer", run_on_startup=False,
               use_monitor=False) 
 def $(FUNCTION_NAME_INPUT)(myTimer: func.TimerRequest) -> None:
     utc_timestamp = datetime.datetime.utcnow().replace(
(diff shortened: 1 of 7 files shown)


Docs/Actions/ReplaceTokensInText.md
```diff
@@ -142,7 +142,7 @@ $(TIMER_FUNCTION_BODY)
 
 ```python
 @app.function_name(name="$(FUNCTION_NAME_INPUT)")
-@app.schedule(schedule="$(SCHEDULE_INPUT)", arg_name="myTimer", run_on_startup=True,
+@app.timer_trigger(schedule="$(SCHEDULE_INPUT)", arg_name="myTimer", run_on_startup=False,
```
surgupta-msft, line 145: Curious to understand the context behind changing `run_on_startup` to `False`?

liliankasem:
@gavin-aguiar

Docs/Actions/ReplaceTokensInText.md
gavin-aguiar: run_on_startup should always be false by default. Its default in V1 so one reason is for parity. But main reasons why this should always be false by default is for production, we don't want customers to accidentally create too many timers that run on startup. Updated the description with more details.

PR #1478, opened 2023-10-24
Update templates to use asp.net integration

### LINKED ISSUE #1465: Create HTTP templates using ASP.NET integration
Create a set of templates using the ASP.NET integration packages and programming model

The initial set of templates should include:
- [ ] Basic HTTP triggered function (similar to in-proc template)
- [ ] HTTP triggered function showing model binding support

satvu:
Part of work for [1465](https://github.com/Azure/azure-functions-dotnet-worker/issues/1465) in dotnet-worker repo.

Official guidance [here](https://learn.microsoft.com/en-us/azure/azure-functions/dotnet-isolated-process-guide#aspnet-core-integration)

### DIFF
--- a/Functions.Templates/Templates/HttpTrigger-CSharp-Isolated/.template.config/template.json
+++ b/Functions.Templates/Templates/HttpTrigger-CSharp-Isolated/.template.config/template.json
@@ -15,37 +15,18 @@
     },
     "sourceName": "HttpTriggerCSharp",
     "symbols": {
-        "Framework": {
-            "type": "parameter",
-            "description": "The target framework for the project.",
-            "datatype": "choice",
-            "defaultValue": "net6.0",
-            "choices": [
-                {
-                "choice": "net6.0",
-                "description": "Target .NET 6"
-                },
-                {
-                "choice": "net7.0",
-                "description": "Target .NET 7"
-                },
-                {
-                "choice": "net8.0",
-                "description": "Target .NET 8"
-                },
-                {
-                "choice": "net48",
-                "description": "Target .NET Framework 4.8"
-                }
-            ]
+        "FrameworkIdentifier": {
+            "type": "bind",
+            "binding": "msbuild:TargetFrameworkIdentifier",
+            "defaultValue": ".NETCoreApp"
         },
         "NetFramework": {
             "type": "computed",
-            "value": "(Framework == \"net48\")"
+            "value": "(FrameworkIdentifier == \".NETFramework\")"
         },
         "NetCore": {
             "type": "computed",
-            "value": "(Framework == \"net8.0\" || Framework == \"net7.0\" || Framework == \"net6.0\")"
+            "value": "(FrameworkIdentifier == \".NETCoreApp\")"
         },
         "namespace": {
             "description": "namespace for the generated code",
--- a/Functions.Templates/Templates/HttpTrigger-CSharp-Isolated/.template.config/template.json
+++ b/Functions.Templates/Templates/HttpTrigger-CSharp-Isolated/.template.config/template.json
@@ -15,18 +15,18 @@
     },
     "sourceName": "HttpTriggerCSharp",
     "symbols": {
-        "FrameworkIdentifier": {
+        "Framework": {
             "type": "bind",
-            "binding": "msbuild:TargetFrameworkIdentifier",
-            "defaultValue": ".NETCoreApp"
+            "binding": "msbuild:TargetFramework",
+            "defaultValue": "net6.0"
         },
         "NetFramework": {
             "type": "computed",
-            "value": "(FrameworkIdentifier == \".NETFramework\")"
+            "value": "(Framework == \"net48\")"
         },
         "NetCore": {
             "type": "computed",
-            "value": "(FrameworkIdentifier == \".NETCoreApp\")"
+            "value": "(Framework == \"net6.0\" || Framework == \"net7.0\" || Framework == \"net6.0\")"
         },
         "namespace": {
             "description": "namespace for the generated code",
@@ -67,7 +67,7 @@
     "defaultName": "HttpTriggerCSharp",
     "postActions": [
         {
-            "condition": "NetCore",
+            "condition": "(NetCore == \"True\")",
             "Description": "Adding Reference to Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore Nuget package",
             "ActionId": "B17581D1-C5C9-4489-8F0A-004BE667B814",
             "ContinueOnError": "true",
--- a/Functions.Templates/Templates/HttpTrigger-CSharp-Isolated/.template.config/template.json
+++ b/Functions.Templates/Templates/HttpTrigger-CSharp-Isolated/.template.config/template.json
@@ -54,6 +54,19 @@
     "defaultName": "HttpTriggerCSharp",
     "postActions": [
         {
+            "condition": "(TargetFrameworkIdentifier == \".NETCoreApp\")",
+            "Description": "Adding Reference to Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore Nuget package",
+            "ActionId": "B17581D1-C5C9-4489-8F0A-004BE667B814",
+            "ContinueOnError": "true",
+            "ManualInstructions": [],
+            "args": {
+                "referenceType": "package",
+                "reference": "Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore", "version": "1.0.0",
+                "projectFileExtensions": ".csproj"
+            }
+        },
+        {
+            "condition": "(TargetFrameworkIdentifier == \".NETFramework\")",
             "Description": "Adding Reference to Microsoft.Azure.Functions.Worker.Extensions.Http Nuget package",
             "ActionId": "B17581D1-C5C9-4489-8F0A-004BE667B814",
             "ContinueOnError": "true",
@@ -61,17 +74,6 @@
             "args": {
                 "referenceType": "package",
                 "reference": "Microsoft.Azure.Functions.Worker.Extensions.Http", "version": "3.1.0",
-                "projectFileExtensions": ".csproj"
-            }
-        },
-        {
-            "Description": "Adding Reference to Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore Nuget package",
-            "ActionId": "B17581D1-C5C9-4489-8F0A-004BE667B814",
-            "ContinueOnError": "true",
-            "ManualInstructions": [],
-            "args": {
-                "referenceType": "package",
-              "reference": "Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore", "version": "1.0.0",
                 "projectFileExtensions": ".csproj"
             }
         },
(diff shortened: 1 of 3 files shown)


Functions.Templates/ProjectTemplate_v4.x/CSharp-Isolated/Program.cs
```diff
@@ -19,7 +19,16 @@ static void Main(string[] args)
     }
 }
 #endif
-#if NetCore
+#if NET6_0_OR_GREATER
```
satvu, line 22: Do we need this level of granularity? Or can we just keep the original `#if NetCore` conditional? 
([the ext package targets net6](https://github.com/Azure/azure-functions-dotnet-worker/blob/e4f4b4db4902454b35ae18a4f9948b896f7941f6/extensions/Worker.Extensions.Http.AspNetCore/src/Worker.Extensions.Http.AspNetCore.csproj#L10))

Functions.Templates/Templates/HttpTrigger-CSharp-Isolated/.template.config/template.json
```diff
@@ -64,6 +64,17 @@
                 "projectFileExtensions": ".csproj"
             }
         },
+        {
+            "Description": "Adding Reference to Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore Nuget package",
```
satvu, line 68: We only need this step for some apps (non-framework versions). @soninaren Any guidance for how to go about this?

Functions.Templates/Templates/HttpTrigger-CSharp-Isolated/.template.config/template.json
```diff
@@ -64,6 +64,17 @@
...
+            "ActionId": "n/a",
```
satvu, line 69: Not sure where this value comes from ...

Functions.Templates/Templates/HttpTrigger-CSharp-Isolated/HttpTriggerCSharp.cs
```diff
@@ -15,6 +20,12 @@ public HttpTriggerCSharp(ILoggerFactory loggerFactory)
         }
 
         [Function("HttpTriggerCSharp")]
+#if NET6_0_OR_GREATER
```
satvu, line 23: Do we want this level of granularity or is `#if NETCORE` good enough?

Functions.Templates/Templates/HttpTrigger-CSharp-Isolated/.template.config/template.json
soninaren: Action Id is predefined guid from templating engine. You can copy this from another template.

Functions.Templates/Templates/HttpTrigger-CSharp-Isolated/.template.config/template.json
soninaren: Can you elaborate on some apps part?

satvu: This package shouldn't be added for net framework apps, but it seems like this template is shared for both netframework and netcore.

Functions.Templates/Templates/HttpTrigger-CSharp-Isolated/HttpTriggerCSharp.cs
```diff
@@ -1,20 +1,31 @@
...
     public class HttpTriggerCSharp
     {
-        private readonly ILogger _logger;
+        private readonly ILogger<T> _logger;
 
-        public HttpTriggerCSharp(ILoggerFactory loggerFactory)
+        public HttpTriggerCSharp(ILogger<T> logger)
         {
-            _logger = loggerFactory.CreateLogger<HttpTriggerCSharp>();
+            _logger = logger;
```
mattchenderson, line 19: `T` here should be `HttpTriggerCSharp`, right?

Functions.Templates/Templates/HttpTrigger-CSharp-Isolated/HttpTriggerCSharp.cs
```diff
@@ -1,20 +1,31 @@
...
         }
 
         [Function("HttpTriggerCSharp")]
+#if NetCore
+        public IActionResult Run([HttpTrigger(AuthorizationLevel.Anonymous, "get")] HttpRequest req)
```
mattchenderson, line 24: Authorization levels should be the same in both cases. AuthLevelValue is correct here I think because it will be set as a parameter to the template engine.

Functions.Templates/Templates/HttpTrigger-CSharp-Isolated/.template.config/template.json
```diff
@@ -54,6 +54,19 @@
...
     "postActions": [
         {
+            "condition": "(TargetFrameworkIdentifier == \".NETCoreApp\")",
+            "Description": "Adding Reference to Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore Nuget package",
+            "ActionId": "B17581D1-C5C9-4489-8F0A-004BE667B814",
+            "ContinueOnError": "true",
+            "ManualInstructions": [],
+            "args": {
+                "referenceType": "package",
+                "reference": "Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore", "version": "1.0.0",
```
mattchenderson, line 64: This version will need to be updated once release is cut, right?

mattchenderson: Should there also be a framework reference for ASP.NET Core here as well? `<FrameworkReference Include="Microsoft.AspNetCore.App" />`

Functions.Templates/Templates/HttpTrigger-CSharp-Isolated/HttpTriggerCSharp.cs
mattchenderson: Method choice should also be the same, I think.

mattchenderson (changes requested):

Functions.Templates/Templates/HttpTrigger-CSharp-Isolated/.template.config/template.json
satvu: Confused by this ask - what does that mean in the context of a `template.json` file?

mattchenderson: I'm assuming the scenario where the customer has not included this in their project file otherwise. Since it's a requirement here, I'm wondering if it makes sense to add an additional reference. Does `referenceType` allow for a value like "framework" or something?

mattchenderson: It would appear that it does not... https://github.com/dotnet/templating/wiki/Post-Action-Registry#add-a-reference-to-a-project-file
"package" or "project", unfortunately

(Also hey look - in-proc is the first example there)

Functions.Templates/Templates/HttpTrigger-CSharp-Isolated/HttpTriggerCSharp.cs
```diff
@@ -1,20 +1,31 @@
...
         {
-            _logger = loggerFactory.CreateLogger<HttpTriggerCSharp>();
+            _logger = logger;
         }
 
         [Function("HttpTriggerCSharp")]
+#if NetCore
+        public IActionResult Run([HttpTrigger(AuthorizationLevel.AuthLevelValue, "get", "post")] HttpRequest req)
+        {
+            return new OkObjectResult("Welcome to Azure Functions!");
```
mattchenderson, line 26: The logger isn't used in the NetCore path. We should make sure to include that,
```suggestion
            _logger.LogInformation("C# HTTP trigger function processed a request.");

            return new OkObjectResult("Welcome to Azure Functions!");
```

Functions.Templates/Templates/HttpTrigger-CSharp-Isolated/.template.config/template.json
```diff
@@ -53,6 +85,30 @@
...
     "postActions": [
+        {
+            "condition": "NetCore",
+            "Description": "Adding Reference to Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore Nuget package",
+            "ActionId": "B17581D1-C5C9-4489-8F0A-004BE667B814",
+            "ContinueOnError": "true",
+            "ManualInstructions": [],
+            "args": {
+                "referenceType": "package",
+                "reference": "Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore", "version": "1.1.0",
```
fabiocav, line 96: We pushed a version with worker dependency updates and the latest is now `1.1.1`

Functions.Templates/Templates/HttpTrigger-CSharp-Isolated/.template.config/template.json
```diff
@@ -53,6 +85,30 @@
...
+                "projectFileExtensions": ".csproj"
+            }
+        },
+        {
+            "condition": "NetCore",
+            "Description": "Adding Reference to Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore.Analyzers Nuget package",
```
fabiocav, line 102: We shouldn't need an explicit reference here. The extension package should bring this along.

Functions.Templates/Templates/HttpTrigger-CSharp-Isolated/.template.config/template.json
```diff
@@ -15,6 +15,19 @@
...
+            "binding": "msbuild:TargetFramework",
+            "defaultValue": "net6.0"
+        },
+        "NetFramework": {
+            "type": "computed",
+            "value": "(Framework == \"net48\")"
+        },
+        "NetCore": {
+            "type": "computed",
+            "value": "(Framework == \"net6.0\" || Framework == \"net7.0\" || Framework == \"net6.0\")"
```
mattchenderson, line 29:
```suggestion
            "value": "(Framework == \"net6.0\" || Framework == \"net7.0\" || Framework == \"net8.0\")"
```

mattchenderson:
@satvu Following up on our chat earlier today - the testing with the `dotnet` CLI may have been having the odd binding behavior because a `dotnet restore` is needed in order for the `msbuild:TargetFramework` bindings to be hydrated. I think a way to smooth this over would be to force a project restore from the project template.  Here's an example: https://github.com/dotnet/templating/blob/main/dotnet-template-samples/content/08-restore-on-create/MyProject.Con.CSharp/.template.config/template.json

VS should be doing a restore for you automatically, so I don't know how to explain anything seen there, but using the latest build from this PR and adding an explicit `dotnet restore` in, everything seems to be working correctly on my machine using the `dotnet` CLI.

Full command set for .NET Framework:
```console
dotnet new func -v diag -d -F net48
dotnet restore
dotnet new http -v diag -d
```

Full command set for .NET (6 because [8 is not in the current build for this PR](https://github.com/Azure/azure-functions-templates/pull/1478#discussion_r1390020543)):
```console
dotnet new func -v diag -d -F net6.0
dotnet restore
dotnet new http -v diag -d
```

In both cases, the `Program.cs`, `.csproj`, and `.cs` look good***. I see the project file updated appropriately by the item template.

We should do a sniff test with Core Tools and VS (I didn't want to figure out how to set those up with the build drop), but I think the current state seems close. We just need to get .NET 8 included.

*** I'm still of the opinion that the project template should be frontloading this conditionality so that `Program.cs` doesn't need a manual change after this set of commands.

satvu:
Closing this PR in favor of restructuring item templates by framework version instead of combining them into one template (per discussion offline). This version requires changes to VS and would break on past versions.

PR #1480, opened 2023-11-01
Sql vbump

MaddyDev:
Sql vbump for trigger GA.

Fixes https://github.com/Azure/azure-functions-sql-extension/issues/956


lucyzhang929 (approved):

Charles-Gagnon (approved):

Charles-Gagnon:
It looks like this is also fixing https://github.com/Azure/azure-functions-sql-extension/issues/956 too, is that correct?

MaddyDev:
> It looks like this is also fixing [Azure/azure-functions-sql-extension#956](https://github.com/Azure/azure-functions-sql-extension/issues/956) too, is that correct?

yes, I addressed the fix in this as well.

PR #1491, opened 2023-12-01
Add ASP.NET Integration Http Trigger Templates

### LINKED ISSUE #1465: Create HTTP templates using ASP.NET integration

satvu:
Resolves https://github.com/Azure/azure-functions-dotnet-worker/issues/1465


Functions.Templates/Templates/HttpTrigger-CSharp-Isolated-NetFx/.template.config/template.json
```diff
@@ -0,0 +1,84 @@
+{
+    "author": "Microsoft",
+    "classifications": [
+        "Azure Function",
+        "Trigger",
+        "Http"
+    ],
+    "name": "HttpTrigger",
+    "identity": "Azure.Function.CSharp.Isolated.NetFx.HttpTrigger.1.x",
```
satvu, line 9: @soninaren Put a placeholder for now, but I'm wondering if this needs a different name from the other CSharp-Isolated trigger and what the naming convention is for these item templates

mattchenderson: Does defining a new identity at this level imply other segmentation anywhere?

I had kind of assumed that the segmentation here would be in [NuGet packages](https://github.com/Azure/azure-functions-templates/tree/dev/Build/PackageFiles/Dotnet_precompiled) where included files representing "HttpTrigger" differ between them (ideally the definition for the other stuff being in one shared source of truth).

With different packages, I know how we would have VS consume them differently. I'm less certain about other environments, such as Core Tools and VS Code... I also don't know what signal they would be using to disambiguate between the framework versions, either. We may need to add some plumbing there to pull the TFM out of the project file.

Functions.Templates/Templates/HttpTrigger-CSharp-Isolated-NetFx/HttpTriggerCSharp.cs
```diff
@@ -0,0 +1,30 @@
+using System.Net;
+using Microsoft.Azure.Functions.Worker;
+using Microsoft.Azure.Functions.Worker.Http;
+using Microsoft.Extensions.Logging;
```
mattchenderson, line 4: nit: sort
```suggestion
using Microsoft.Azure.Functions.Worker;
using Microsoft.Azure.Functions.Worker.Http;
using Microsoft.Extensions.Logging;
using System.Net;
```

Functions.Templates/Templates/HttpTrigger-CSharp-Isolated/HttpTriggerCSharp.cs
```diff
@@ -1,30 +1,24 @@
-using System.Net;
 using Microsoft.Azure.Functions.Worker;
-using Microsoft.Azure.Functions.Worker.Http;
 using Microsoft.Extensions.Logging;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
```
mattchenderson, line 4: nit: These should be first in sort order

Functions.Templates/Templates/HttpTrigger-CSharp-Isolated-NetFx/.template.config/template.json
mattchenderson: OH core tools DOES have something for getting the TFM. They do it for publish:
https://github.com/Azure/azure-functions-core-tools/blob/2bb506b73fd22e180131957aa631b08176805dc8/src/Azure.Functions.Cli/Actions/AzureActions/PublishFunctionAppAction.cs#L166-L179

I could therefore imagine us doing an explicit "is this .NET Framework" check during the process and swapping which template maps to that shortName or something.

Functions.Templates/Templates/HttpTrigger-CSharp-Isolated-NetFx/.template.config/template.json
```diff
@@ -0,0 +1,84 @@
...
+    "groupIdentity": "Azure.Function.HttpTrigger",
+    "shortName": "http",
```
satvu, line 11: note: will need a different short name

satvu: Closing - only need different short name if there are two triggers in one package but we are going with splitting.

satvu: Jk core-tools will ship both packages together so changing this

Build/PackageFiles/Dotnet_precompiled/ItemTemplates-Isolated-NetFx_v4.x.nuspec
```diff
@@ -0,0 +1,119 @@
...
+    <file src="..\..\..\Functions.Templates\Templates\QueueTrigger-CSharp-Isolated\.template.config\vs-2017.3\*.*" target="content\QueueTrigger-CSharp\.template.config\vs-2017.3" />
+    <file src="..\..\..\Functions.Templates\Templates\QueueTrigger-FSharp-Isolated\QueueTriggerFSharp.fs" target="content\QueueTrigger-FSharp\QueueTriggerFSharp.fs" />
+    <file src="..\..\..\Functions.Templates\Templates\QueueTrigger-FSharp-Isolated\.template.config\template.json" target="content\QueueTrigger-FSharp\.template.config\template.json" />
+    <file src="..\..\..\Functions.Templates\Templates\QueueTrigger-FSharp-Isolated\.template.config\vs-2017.3.host.json" target="content\QueueTrigger-FSharp\.template.config\vs-2017.3.host.json" />
+    <file src="..\..\..\Functions.Templates\Templates\QueueTrigger-FSharp-Isolated\.template.config\vs-2017.3\*.*" target="content\QueueTrigger-FSharp\.template.config\vs-2017.3" />
+    <file src="..\..\..\Functions.Templates\Templates\HttpTrigger-CSharp-Isolated-NetFx\HttpTriggerCSharp.cs" target="content\HttpTrigger-CSharp\HttpTriggerCSharp.cs" />
```
satvu, line 27: All templates are the same except this one - points to `\HttpTrigger-CSharp-Isolated-NetFx\*` instead

Build/PackageFiles/Dotnet_precompiled/ItemTemplates-Isolated-NetFx_v4.x.nuspec
```diff
@@ -0,0 +1,119 @@
+<?xml version="1.0"?>
+<package >
+  <metadata>
+    <id>Microsoft.Azure.Functions.Worker.NetFx.ItemTemplates</id>
```
satvu, line 4: Different id

Build/PackageFiles/Dotnet_precompiled/ItemTemplates-Isolated-NetFx_v4.x.nuspec
```diff
@@ -0,0 +1,119 @@
...
+    <copyright>© Microsoft Corporation. All rights reserved.</copyright>
+    <requireLicenseAcceptance>true</requireLicenseAcceptance>
+    <description>Collection of templates for Azure Functions .NET Isolated using .NET Framework</description>
+    <language>en-US</language>
+    <packageTypes>
+      <packageType name="Template" />
+    </packageTypes>
+  </metadata>
+  <files>
+    <file src="..\..\..\Functions.Templates\Templates\QueueTrigger-CSharp-Isolated\QueueTriggerCSharp.cs" target="content\QueueTrigger-CSharp\QueueTriggerCSharp.cs" />
```
soninaren, line 19: Can likely follow the content format from this file. Easier to manage and review.
https://github.com/Azure/azure-functions-templates/blob/dev/Build/PackageFiles/Dotnet_precompiled/ItemTemplates-Isolated_v3.x.nuspec

satvu: Shortened it to use the V3 content format, but would love a second pair of eyes to confirm that this new version still has all of the triggers present from https://github.com/Azure/azure-functions-templates/blob/dev/Build/PackageFiles/Dotnet_precompiled/ItemTemplates-Isolated_v4.x.nuspec

Build/PackageFiles/Dotnet_precompiled/ItemTemplates-Isolated-NetFx_v4.x.nuspec
```diff
@@ -0,0 +1,48 @@
...
+    <language>en-US</language>
+    <packageTypes>
+      <packageType name="Template" />
+    </packageTypes>
+  </metadata>
+  <files>
+    <file src="..\..\..\Functions.Templates\Templates\QueueTrigger-CSharp-Isolated\**" target="content\QueueTrigger-CSharp\QueueTriggerCSharp.cs" />
+    <file src="..\..\..\Functions.Templates\Templates\QueueTrigger-FSharp-Isolated\**" target="content\QueueTrigger-FSharp\QueueTriggerFSharp.fs" />
+    <file src="..\..\..\Functions.Templates\Templates\HttpTrigger-CSharp-Isolated-NetFx\**" target="content\HttpTrigger-CSharp\HttpTriggerCSharp.cs" />
+    <file src="..\..\..\Functions.Templates\Templates\HttpTrigger-FSharp-Isolated\**" target="content\HttpTrigger-FSharp\HttpTriggerFSharp.fs" />
```
mattchenderson, line 22: `HttpTrigger-FSharp-Isolated` theoretically would also need to be updated, but I suppose that depends on what happens with the project template per another PR discussion. I'm ok keeping this out of the scope of this one, but I just wanted to note it.

Functions.Templates/Templates/HttpTrigger-CSharp-Isolated-NetFx/.template.config/template.json
mattchenderson: Wait, so what happens if someone wants to just use the `dotnet` CLI? I guess they too would install the netfx version, but then how do they tell the system which to use? Am I correct in assuming that the Core Tools is only doing the changeup for HTTP, or would we do it across the board for all templates?

Functions.Templates/Templates/HttpTrigger-CSharp-Isolated-NetFx/.template.config/template.json
mattchenderson: (Related to this question: https://github.com/Azure/azure-functions-templates/pull/1491/files#r1421185758)

If we're doing the short name mapping, I wonder if that is just something that happens for HTTP (in which case Core Tools would just need the one package, that including the new template with an alternative short name). I'm not clear on where the selection of which template package to use would occur otherwise.

Functions.Templates/Templates/HttpTrigger-CSharp-Isolated-NetFx/.template.config/template.json
soninaren: Given that the rest (not http trigger) of the templates are same, it might be okay to pick either one. But for http we do need a different short name if both the packages are to be installed and used at the same time. I am not sure how dotnet CLI handles the cache for two templates with same short name. We would have to test this out to see how it works.

mattchenderson:
Do we need to add a restore step on the project templates? It won't work for VS, but that presumably would be needed for other clients that rely on the msbuild context from the existing package, I think?

liliankasem (approved):

satvu:
Closing in favor of #1495

PR #1495, opened 2023-12-20
[ASP.NET Core Integration] Add templates using 3-package approach

satvu:
Resolves https://github.com/Azure/azure-functions-dotnet-worker/issues/1465

Validations:

- [x] Dotnet CLI 
  - [x] NetFx
  - [x] NetCore
 - [x] Core Tools (following [this](https://learn.microsoft.com/en-us/azure/azure-functions/functions-run-local?tabs=windows%2Cisolated-process%2Cnode-v4%2Cpython-v2%2Chttp-trigger%2Ccontainer-apps&pivots=programming-language-csharp#create-func) flow)
   - [x] NetFx
   - [x] NetCore

VS validation will have to be done with a tooling-feed test to make sure NetFx and NetCore scenarios point to the correct package.

## Additional Notes

- If you use the Core package in a net48 project context, it spits out the core version but the nuget install seems to fail silently - we do NOT expect this pattern as customers should not be mixing templates for different frameworks when using CLI or core-tools. This will be controlled on our end for VS.
  - The inverse of this also leads to failure (same error as in https://github.com/Azure/azure-functions-dotnet-worker/issues/2184), but that is also an unsupported/unexpected pattern.


Build/PackageFiles/Dotnet_precompiled/ItemTemplates-Isolated_NetCoreVS_v4.x.nuspec
```diff
@@ -0,0 +1,50 @@
+<?xml version="1.0"?>
+<package >
+  <metadata>
+    <id>Microsoft.Azure.Functions.Worker.ItemTemplates.NetCoreVS</id>
```
satvu, line 4: Thoughts on name + versioning for new package? Would also like reviews for the other newly introduced package.

Functions.Templates/ProjectTemplate_v4.x/CSharp-Isolated/.template.config/template.json
```diff
@@ -101,5 +101,38 @@
             "path": "Company.FunctionApp.csproj"
         }
     ],
-    "defaultName": "Company.FunctionApp"
-}
+    "defaultName": "Company.FunctionApp",
+    "postActions":
+    [
+        {
+            "condition": "(NetCore)",
```
satvu, line 108: Need to add these packages to the post action for the asp.net wireup

Functions.Templates/ProjectTemplate_v4.x/CSharp-Isolated/Program.cs
```diff
@@ -30,11 +30,7 @@ static void Main(string[] args)
 using Microsoft.Extensions.DependencyInjection;
 
 var host = new HostBuilder()
-    .ConfigureFunctionsWorkerDefaults()
-    .ConfigureServices(services => {
-        services.AddApplicationInsightsTelemetryWorkerService();
-        services.ConfigureFunctionsApplicationInsights();
-    })
+    .ConfigureFunctionsWebApplication()
```
satvu, line 33: Discussed offline: may be better to pull in unnecessary dependencies for non-http apps than to have customers trying out ASP.NET integration entering a broken state.

mattchenderson: We should still wire up App Insights though
```suggestion
    .ConfigureFunctionsWebApplication()
    .ConfigureServices(services => {
        services.AddApplicationInsightsTelemetryWorkerService();
        services.ConfigureFunctionsApplicationInsights();
    })
```

mattchenderson: wait actually isn't this a file that needs a conditional based on netcore overall? It no longer serves both types of targets. So that suggestion in my previous comment is off. But I thought this file used to be conditional to account for the debugger enablement anyway?

mattchenderson (changes requested):
Just requesting changes to Program.cs to reintroduce the logging wire-up per other comment. Otherwise testing from this branch, all behaviors are as I would expect for both the Core Tools and the .NET CLI.

mattchenderson:
Also validated in VS Code, so I'm feeling good about this set of changes once the App Insights wire-up is back in there.

mattchenderson (approved):

fabiocav (approved):

Build/PackageFiles/Dotnet_precompiled/ItemTemplates-Isolated_NetCoreVS_v4.x.nuspec
```diff
@@ -0,0 +1,50 @@
...
+    <requireLicenseAcceptance>true</requireLicenseAcceptance>
+    <description>Collection of templates for Azure Functions .NET Isolated</description>
+    <language>en-US</language>
+    <packageTypes>
+      <packageType name="Template" />
+    </packageTypes>
+  </metadata>
+  <files>
+    <!-- Target unique HTTP package for VS feed-->
+    <file src="..\..\..\Functions.Templates\Templates\HttpTrigger-CSharp-Isolated-NetCoreVS\**" target="content\HttpTrigger-CSharp" />
```
soninaren, line 20:
```suggestion
    <file src="..\..\..\Functions.Templates\Templates\HttpTrigger-CSharp-Isolated-NetCore\**" target="content\HttpTrigger-CSharp" />
```

Build/PackageFiles/Dotnet_precompiled/ItemTemplates-Isolated_NetFxVS_v4.x.nuspec
```diff
@@ -0,0 +1,49 @@
...
+    <requireLicenseAcceptance>true</requireLicenseAcceptance>
+    <description>Collection of templates for Azure Functions .NET Isolated</description>
+    <language>en-US</language>
+    <packageTypes>
+      <packageType name="Template" />
+    </packageTypes>
+  </metadata>
+  <files>
+    <!-- Target unique HTTP package for VS feed-->
+    <file src="..\..\..\Functions.Templates\Templates\HttpTrigger-CSharp-Isolated-NetFxVS\**" target="content\HttpTrigger-CSharp" />
```
soninaren, line 20:
```suggestion
    <file src="..\..\..\Functions.Templates\Templates\HttpTrigger-CSharp-Isolated-NetFx\**" target="content\HttpTrigger-CSharp" />
```

Build/PackageFiles/Dotnet_precompiled/ItemTemplates-Isolated_NetFxVS_v4.x.nuspec
```diff
@@ -0,0 +1,49 @@
+<?xml version="1.0"?>
+<package >
+  <metadata>
+    <id>Microsoft.Azure.Functions.Worker.ItemTemplates.NetFxVS</id>
```
soninaren, line 4:
```suggestion
    <id>Microsoft.Azure.Functions.Worker.ItemTemplates.NetFx</id>
```

Functions.Templates/Templates/HttpTrigger-CSharp-Isolated-NetCoreVS/.template.config/template.json
```diff
@@ -0,0 +1,95 @@
+{
+    "author": "Microsoft",
+    "classifications": [
+        "Azure Function",
+        "Trigger",
+        "Http"
+    ],
+    "name": "HttpTrigger",
+    "identity": "Azure.Function.CSharp.Isolated.HttpTrigger.NetCoreVS.1.x",
```
soninaren, line 9:
```suggestion
    "identity": "Azure.Function.CSharp.Isolated.HttpTrigger.NetCore.1.x",
```

soninaren (approved):
Nit comments on package and directory renames. Looks good otherwise.

PR #1529, opened 2024-05-02
updating in-process proj template to support net8

mattchenderson:
Adding flexibility to the in-process template to support other TFMs. Draft PR while we await an appropriate version of the build SDK (4.4.0 does not exist at the time of this PR being opened).


fabiocav (approved):

fabiocav:
Merging this. We can follow up if required by additional SDK changes.

PR #1519, opened 2024-04-18
Add Event Grid Blob Storage trigger templates

castrodd:
This PR is adding templates for the Event Grid-based Blob Storage trigger (as described in [this tutorial](https://learn.microsoft.com/en-us/azure/azure-functions/functions-event-grid-blob-trigger?tabs=isolated-process%2Cnodejs-v4&pivots=programming-language-csharp)).

What's included:
- Templates for C# (isolated), Python, JavaScript/TypeScript (old model), and PowerShell

What's _not_ included:
- Templates for in-process C#, F#, Java, JavaScript/TypeScript (new model)
    - In-process C# not supported on Flex, so deprioritizing those templates
    - F# is not listed in the above tutorial, so not including in this first wave of sample templates
    - Java templates are not included in this repository
      - Java template also does not seem to be working. Still need to verify it actually works. 
    - New model JS/TS templates are not included in this repository

PR #1534, opened 2024-05-20
Add EventGrid Blob trigger for Python v2

castrodd:
This PR is a follow-up to #1519.

In that PR, we included the Event Grid Blob trigger for Python v1, but we failed to include it for Python v2.

This PR addresses that oversight.


hallvictoria (approved):

paulyuk:
Thank you @castrodd 

Note to see this fix you need to update your `requirements.txt` package to the latest, e.g.
```text
azure-functions>=1.21.3
```

I have a full sample that leverages EventGrid triggered BlobTrigger here, and it deploys to Flex Consumption
https://github.com/Azure-Samples/function-python-ai-textsummarize

PR #1550, opened 2024-07-23
Updating build for 1es migration

soninaren:


eng/official-build.yml
```diff
@@ -0,0 +1,49 @@
...
+      image: 1es-windows-2022
+      os: windows
+
+    stages:  
+    - stage: Build
+      dependsOn: []
+      jobs:
+      - template: eng/ci/templates/jobs/build.yml@self
+        parameters:
+          official: true
```
jviau: there is no `official` parameter declared by `build.yml`

eng/public-build.yml
```diff
@@ -0,0 +1,32 @@
...
+  template: v1/1ES.Unofficial.PipelineTemplate.yml@1es
+  parameters:
+    pool:
+      name: 1es-pool-azfunc-public
+      image: 1es-windows-2022
+      os: windows
+
+    stages:
+    - stage: Build
+      dependsOn: []
```
jviau, line 30: This can be removed

eng/official-build.yml
```diff
@@ -0,0 +1,49 @@
...
+extends:
+  template: v1/1ES.Official.PipelineTemplate.yml@1es
+  parameters:
+    pool:
+      name: 1es-pool-azfunc
+      image: 1es-windows-2022
+      os: windows
+
+    stages:  
+    - stage: Build
```
jviau, line 44: are you missing the call to `localization.yml` template?

eng/official-build.yml
soninaren: Going to delete this parameter for now. Will use it to disable localization from public builds as discussed.

eng/official-build.yml
soninaren: Have skipped it on purpose as it is broken. Will be adding it back in another PR once it is fixed

jviau (approved):

PR #1574, opened 2024-10-04
Use IHostApplicationBuilder for .NET Worker V2 SDK

### LINKED ISSUE #1573: Update .NET isolated templates for IHostApplicationBuilder
For .NET 8 and .NET 9 on the isolated worker model, instead of using the generic `IHostBuilder`, we should use [whatever types the worker adds](https://github.com/Azure/azure-functions-dotnet-worker/issues/2745) following the `IHostApplicationBuilder` model.

mattchenderson:
Resolves #1573

Draft PR as the versions which supply the `IHostApplicationBuilder` support are not yet available. The current version reflects my current understanding of the surface area that will be exposed.

The approach taken here is to add a `FrameworkShouldUseV1Sdk` condition over the existing content. If that condition is failed, we instead use the V2 versions. Since V2 is in preview, only .NET 9 (also preview) is excluded from `FrameworkShouldUseV1Sdk`. When V2 is GA'd, .NET 8 and .NET Framework should be removed from `FrameworkShouldUseV1Sdk`, and for .NET Framework, additional changes will need to be made for to its section in `Program.cs`.

### DIFF
--- a/Functions.Templates/ProjectTemplate_v4.x/CSharp-Isolated/.template.config/template.json
+++ b/Functions.Templates/ProjectTemplate_v4.x/CSharp-Isolated/.template.config/template.json
@@ -51,7 +51,11 @@
         },
         "NetCore": {
             "type": "computed",
-            "value": "(Framework == \"net9.0\" || Framework == \"net8.0\"|| Framework == \"net7.0\" || Framework == \"net6.0\")"
+            "value": "(Framework == \"net9.0\" || Framework == \"net8.0\" || Framework == \"net7.0\" || Framework == \"net6.0\")"
+        },
+        "FrameworkShouldUseV1Dependencies": {
+            "type": "computed",
+            "value": "(Framework == \"net48\" || Framework == \"net8.0\" || Framework == \"net7.0\" || Framework == \"net6.0\")"
         },
         "StorageConnectionStringValue": {
             "description": "The connection string for your Azure WebJobs Storage.",
@@ -109,14 +113,26 @@
     "postActions":
     [
         {
-            "condition": "(NetCore)",
+            "condition": "(NetCore && FrameworkShouldUseV1Dependencies)",
             "Description": "Adding Reference to Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore Nuget package",
             "ActionId": "B17581D1-C5C9-4489-8F0A-004BE667B814",
             "ContinueOnError": "true",
             "ManualInstructions": [],
             "args": {
                 "referenceType": "package",
-                "reference": "Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore", "version": "1.2.0",
+                "reference": "Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore", "version": "1.3.2",
+                "projectFileExtensions": ".csproj"
+            }
+        },
+        {
+            "condition": "(NetCore && !FrameworkShouldUseV1Dependencies)",
+            "Description": "Adding Reference to Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore Nuget package",
+            "ActionId": "B17581D1-C5C9-4489-8F0A-004BE667B814",
+            "ContinueOnError": "true",
+            "ManualInstructions": [],
+            "args": {
+                "referenceType": "package",
+                "reference": "Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore", "version": "2.0.0-preview2",
                 "projectFileExtensions": ".csproj"
             }
         },
@@ -128,7 +144,7 @@
             "ManualInstructions": [],
             "args": {
                 "referenceType": "package",
-                "reference": "Microsoft.Azure.Functions.Worker.Extensions.Http", "version": "3.1.0",
+                "reference": "Microsoft.Azure.Functions.Worker.Extensions.Http", "version": "3.2.0",
                 "projectFileExtensions": ".csproj"
             }
         },
--- a/Functions.Templates/ProjectTemplate_v4.x/CSharp-Isolated/Company.FunctionApp.csproj
+++ b/Functions.Templates/ProjectTemplate_v4.x/CSharp-Isolated/Company.FunctionApp.csproj
@@ -13,15 +13,16 @@
   <!--#if (NetCore)-->
     <FrameworkReference Include="Microsoft.AspNetCore.App" />
   <!--#endif -->
-  <!--#if (Framework == "net9.0") -->
-    <PackageReference Include="Microsoft.Azure.Functions.Worker" Version="2.0.0-preview1" />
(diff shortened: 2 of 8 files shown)


mattchenderson:
The ASP.NET Core integration package inclusion (which is a set of postActions) will also need to be updated accordingly. This may also impact item templates for that reason.

Edit: This is done for the project files, and I've updated the F# version to better mirror the C# one (including ASP.NET Core integration, etc.). I also realized the package reference can simply jump up to latest GA for item templates. And we only need to care about the postAction distinction while in the preview period, as that will also be able to move forward. This is effectively the same as the `.csproj` and `.fsproj` changes, which are also really just point in time.

Functions.Templates/ProjectTemplate_v4.x/FSharp-Isolated/.template.config/template.json
```diff
@@ -75,5 +79,50 @@
...
+        },
+        {
+            "condition": "(NetCore)",
+            "Description": "Adding Reference to Microsoft.Azure.Functions.Worker.Extensions.Http Nuget package",
+            "ActionId": "B17581D1-C5C9-4489-8F0A-004BE667B814",
+            "ContinueOnError": "true",
+            "ManualInstructions": [],
+            "args": {
+                "referenceType": "package",
+                "reference": "Microsoft.Azure.Functions.Worker.Extensions.Http", "version": "3.1.0",
```
liliankasem: latest version is `3.2.0`

Functions.Templates/ProjectTemplate_v4.x/CSharp-Isolated/.template.config/template.json
```diff
@@ -109,14 +113,26 @@
...
             "ActionId": "B17581D1-C5C9-4489-8F0A-004BE667B814",
             "ContinueOnError": "true",
             "ManualInstructions": [],
             "args": {
                 "referenceType": "package",
-                "reference": "Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore", "version": "1.2.0",
+                "reference": "Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore", "version": "2.0.0-preview1",
                 "projectFileExtensions": ".csproj"
             }
         },
```
liliankasem, line 138: Update `Microsoft.Azure.Functions.Worker.Extensions.Http` to `"version": "3.2.0"`

Functions.Templates/ProjectTemplate_v4.x/CSharp-Isolated/Company.FunctionApp.csproj
```diff
@@ -13,12 +13,12 @@
...
-  <!--#else-->
+  <!--#if (FrameworkShouldUseV1Sdk) -->
     <PackageReference Include="Microsoft.Azure.Functions.Worker" Version="1.23.0" />
     <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="1.17.4" />
+  <!--#else-->
+    <PackageReference Include="Microsoft.Azure.Functions.Worker" Version="2.0.0-preview1" />
+    <PackageReference Include="Microsoft.Azure.Functions.Worker.Sdk" Version="2.0.0-preview2" />
   <!--#endif -->
     <PackageReference Include="Microsoft.ApplicationInsights.WorkerService" Version="2.22.0" />
     <PackageReference Include="Microsoft.Azure.Functions.Worker.ApplicationInsights" Version="1.2.0" />
```
liliankasem: Should we move this into the if-else brace so that we can load the v1 and v2 i.e.

If FrameworkShouldUseV1Sdk:

`<PackageReference Include="Microsoft.Azure.Functions.Worker.ApplicationInsights" Version="1.4.0" />`

> and update 1.2.0 to 1.4.0

Else:

`<PackageReference Include="Microsoft.Azure.Functions.Worker.ApplicationInsights" Version="2.0.0" />`

Same for the fsharp project

mattchenderson:
Additional changes were needed to prevent item templates from performing a package downgrade.

Functions.Templates/Templates/HttpTrigger-CSharp-Isolated-NetCore/.template.config/template.json
```diff
@@ -73,7 +73,6 @@
             "args": {
                 "referenceType": "package",
                 "reference": "Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore",
-                "version": "1.3.2",
```
mattchenderson, line 76: This removal is intentional and is required to prevent a package downgrade. If a reference is already in place, it will be preserved. If it is not, the system will grab latest stable.

brettsam (approved):

Functions.Templates/ProjectTemplate_v4.x/CSharp-Isolated/.template.config/template.json
```diff
@@ -51,7 +51,11 @@
...
         "NetCore": {
             "type": "computed",
-            "value": "(Framework == \"net9.0\" || Framework == \"net8.0\"|| Framework == \"net7.0\" || Framework == \"net6.0\")"
+            "value": "(Framework == \"net9.0\" || Framework == \"net8.0\" || Framework == \"net7.0\" || Framework == \"net6.0\")"
+        },
+        "FrameworkShouldUseV1Sdk": {
```
brettsam: Nit... it's not really the `V1Sdk` as it's a range of packages right? Should it just be `FrameworkShouldUseV1Worker` or something? Or just `...ShouldUseV1`? 

Not a big deal either way... just jumped out.

Functions.Templates/ProjectTemplate_v4.x/FSharp-Isolated/.template.config/template.json
```diff
@@ -75,5 +79,50 @@
...
+    [
+        {
+            "condition": "(NetCore && FrameworkShouldUseV1Dependencies)",
+            "Description": "Adding Reference to Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore Nuget package",
+            "ActionId": "B17581D1-C5C9-4489-8F0A-004BE667B814",
+            "ContinueOnError": "true",
+            "ManualInstructions": [],
+            "args": {
+                "referenceType": "package",
+                "reference": "Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore", "version": "1.3.2",
```
jviau: Is this formatting intentional? Or should it be separate lines?

```suggestion
                "reference": "Microsoft.Azure.Functions.Worker.Extensions.Http.AspNetCore",
                "version": "1.3.2",
```

Same question for the other lines in this file

jviau (approved):

Functions.Templates/ProjectTemplate_v4.x/FSharp-Isolated/.template.config/template.json
mattchenderson: Oh whoops we had that before but it should indeed have been changed. Will fix now.

PR #1634, opened 2025-03-14
bump npm packages for Build folder

manvkaur:
- Bumping some npm packages to newer versions (some of the major versions/latest versions can't be used due to build issues and need careful review)
- Bumping major for @nestjs/platform-express and gulp-zip is breaking build - this needs to be reviewed and done carefully
- Update build.yml to use node 20


Build/gulp-utils/gulp-resx-js/index.js
```diff
@@ -25,7 +25,7 @@ module.exports = function(opt) {
 
     if (file.isBuffer()) {
       return doConvert(file).then(json => {
-        file.contents = new Buffer(json);
+        file.contents = Buffer.from(json);
```
gavin-aguiar: I'm guessing this change is needed because of some library update?

gavin-aguiar (approved):

Build/gulp-utils/gulp-resx-js/index.js
manvkaur: I was getting a warning for this method during build, but that warning doesn't go away even after introducing this change

PR #1680, opened 2025-07-22
Add CodeQL.yml to try and suppress the scan

jviau:
CodeQL is currently expecting a C# report since it sees `.cs` files. Given these are templates, there are never actually built. Adding this suppression file to see if it will prevent CodeQL from expecting the report.

### DIFF
--- a/CodeQL.yml
+++ b/CodeQL.yml
@@ -1,0 +1,5 @@
+path_classifiers:
+  # While not 'generated' in the traditional sense, these files are templates and will be 'generated'
+  # at template creation time for the customer.
+  generated:
+    - Functions.Templates/**


CodeQL.yml
```diff
@@ -0,0 +1,5 @@
+path_classifiers:
+  # While not 'generated' in the traditional sense, these files are templates and will be 'generated'
+  # at template creation time for the customer.
+  generated:
+    - Functions.Templates/**
```
manvkaur, line 5: Can this be consumed by 1ES PT?

As per my knowledge of 1ESPT, we can specify codeql languages in the official build using following stage variables:

```
variables:
          Codeql.Language: csharp,javascript,powershell,python
          Codeql.RunSourceLanguagesInSourceAnalysis: true
```

manvkaur: variable to exclude certain paths from scan

```
- name: codeql.excludePathPatterns
    value: deps/,build/
```

jviau: Yes, `CodeQL.yml` is used in 1ESPT. We use it in the functions host: https://github.com/Azure/azure-functions-host/blob/dev/CodeQL.yml

manvkaur (approved):

jviau:
Thanks @manvkaur. I will check with the CodeQL team to see if this resolves the issue. Trying to avoid having to do anything drastic in this repo to appease codeql here.

PR #1684, opened 2025-08-06
Updating project templates to support .NET 10

### LINKED ISSUE #1679: Add support for .NET 10 in project templates
.NET 10 needs to be added as a valid framework selection for the project template, and we should conditionally bump the package versions as appropriate, too. The SDK can be updated across the board on the 2.x major, and this could be done now.

The worker package should be conditioned on the TFM for now, as this will be a preview package until .NET 10 GAs. That package release is tracked by https://github.com/Azure/azure-functions-dotnet-worker/issues/3114. Changes here can be staged based on the stated plan, but they shouldn't be merged and released until that package is available through nuget.org.

mattchenderson:
Fixes #1679

Introducing .NET 10 as a target framework and updating package references to support that (which includes a general bump outside of .NET 10).

While testing, I noticed issues with F# that I can reproduce without these changes as well. I will file a separate issue for those.

### DIFF
--- a/Functions.Templates/ProjectTemplate_v4.x/CSharp-Isolated/.template.config/template.json
+++ b/Functions.Templates/ProjectTemplate_v4.x/CSharp-Isolated/.template.config/template.json
@@ -40,6 +40,10 @@
                 "description": "Target .NET 9"
                 },
                 {
+                "choice": "net10.0",
+                "description": "Target .NET 10 (Preview)"
+                },
+                {
                 "choice": "net48",
                 "description": "Target .NET Framework 4.8"
                 }
@@ -51,7 +55,7 @@
         },
         "NetCore": {
             "type": "computed",
-            "value": "(Framework == \"net9.0\" || Framework == \"net8.0\" || Framework == \"net7.0\" || Framework == \"net6.0\")"
+            "value": "(Framework == \"net10.0\" || Framework == \"net9.0\" || Framework == \"net8.0\" || Framework == \"net7.0\" || Framework == \"net6.0\")"
         },
         "FrameworkShouldUseV1Dependencies": {
             "type": "computed",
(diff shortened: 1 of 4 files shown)


Functions.Templates/ProjectTemplate_v4.x/CSharp-Isolated/.template.config/template.json
```diff
@@ -103,6 +107,10 @@
                         "condition": "FrameworkShouldUseV1Dependencies",
                         "value": "1.24.0"
                     },
+                    {
+                        "condition": "FrameworkShouldUseV1Dependencies",
```
kshyju: Do we need the `"condition": "FrameworkShouldUseV1Dependencies"` here? It looks like `FrameworkShouldUseV1Dependencies`  does not have .NET10 (or 8 or 9) in the expression it is evaluating.


https://github.com/Azure/azure-functions-templates/blob/7336d56aae18f119178611d01313ab15290b8eac/Functions.Templates/ProjectTemplate_v4.x/CSharp-Isolated/.template.config/template.json#L56-L59

mattchenderson: That's a bug on my part. Will fix.

kshyju (approved):

manvkaur (approved):

Functions.Templates/Templates/HttpTrigger-FSharp-Isolated/.template.config/template.json
```diff
@@ -56,7 +56,7 @@
         "referenceType": "package",
         "reference": "Microsoft.Azure.Functions.Worker.Extensions.Http",
         "version": "3.3.0",
-        "projectFileExtensions": ".csproj"
+        "projectFileExtensions": ".fsproj"
```
liliankasem, line 59: Yay

PR #1693, opened 2025-09-19
Add template for McpToolTrigger for dotnet-isolated

### LINKED ISSUE #1690: .NET McpToolTrigger template
This should be a simple hello world function:

```csharp
public class HelloTool
{
    private ILogger<HelloTool> _logger;

    public HelloTool(ILogger<HelloTool> logger)
    {
        _logger = logger;
    }

    [Function(nameof(HelloTool))]
    public string Run(
        [McpToolTrigger(nameof(HelloTool), "Responds to the user with a hello message.")] ToolInvocationContext context,
        [McpToolProperty(nameof(name), "string", "The name of the person to greet.")] string? name
    )
    {
        _logger.LogInformation("C# MCP tool trigger function processed a request.");
        return $"Hello, {name ?? "world"}! This is an MCP Tool!";
    }
}
```

liliankasem:
Add template for McpToolTrigger for dotnet-isolated. Validated E2E via Visual Studio.

resolves #1690


Functions.Templates/Templates/McpToolTrigger-CSharp-Isolated/.template.config/template.json
```diff
@@ -0,0 +1,60 @@
+{
+  "author": "Microsoft",
+  "classifications": [
+    "Azure Function",
+    "Trigger",
+    "McpTool" // Or just Mcp?
```
fabiocav: `McpTool` makes sense

Functions.Templates/Templates/McpToolTrigger-CSharp-Isolated/.template.config/template.json
```diff
@@ -0,0 +1,60 @@
+{
+  "author": "Microsoft",
+  "classifications": [
+    "Azure Function",
+    "Trigger",
+    "McpTool" // Or just Mcp?
+  ],
+  "name": "McpToolTrigger",
+  "identity": "Azure.Function.CSharp.Isolated.McpToolTrigger.3.x", // Do we need the .3.x?
```
fabiocav: Why 3.x?

Functions.Templates/Templates/McpToolTrigger-CSharp-Isolated/.template.config/template.json
```diff
@@ -0,0 +1,60 @@
...
+  "postActions": [
+    {
+      "Description": "Adding reference to Microsoft.Azure.Functions.Worker.Extensions.Mcp Nuget package",
+      "ActionId": "B17581D1-C5C9-4489-8F0A-004BE667B814",
+      "ContinueOnError": "true",
+      "ManualInstructions": [],
+      "args": {
+        "referenceType": "package",
+        "reference": "Microsoft.Azure.Functions.Worker.Extensions.Mcp",
+        "version": "1.0.0-preview.7", // update this later for GA once 1.0.0 is out?
```
fabiocav: For this to be merged, it would need to be an existing version.

Functions.Templates/Templates/McpToolTrigger-CSharp-Isolated/.template.config/template.json
liliankasem: I know there was some weird stuff about versioning and names with visual studio but I don't remember the full context. Given this is a brand new template, I wonder if `Azure.Function.CSharp.McpToolTrigger` is okay (without even Isolated?) 

cc: @soninaren

Functions.Templates/Templates/McpToolTrigger-CSharp-Isolated/.template.config/vs-2017.3.host.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": {
+    "text": "MCP Tool trigger",
+    "package": "{E9BCFE2B-304A-4D7F-836F-F3A4FB1A1451}",
+    "id": "8000"
```
liliankasem, line 5: Set to 8000 because it seemed like the an ID that was not used for our existing templates, but I don't know if this is right?

Functions.Templates/Templates/McpToolTrigger-CSharp-Isolated/.template.config/vs-2017.3.host.json
```diff
@@ -0,0 +1,18 @@
...
+  },
+  "description": {
+    "text": "A C# function that will be run whenever it receives an MCP Tool invocation",
+    "package": "{E9BCFE2B-304A-4D7F-836F-F3A4FB1A1451}",
+    "id": "8001"
+  },
+  "order": 400,
```
liliankasem: @soninaren any idea what this is?

Functions.Templates/Templates/McpToolTrigger-CSharp-Isolated/.template.config/template.json
soninaren: 3.x was used to differentiate between runtime versions to avoid duplicate name. But then VS added some hardcoded values setting those identities versions in stone.

We don't need 3.x here :)

Functions.Templates/Templates/McpToolTrigger-CSharp-Isolated/.template.config/vs-2017.3.host.json
soninaren: Might want to change this to bring this template up in the list.

Functions.Templates/Templates/McpToolTrigger-CSharp-Isolated/.template.config/vs-2017.3.host.json
soninaren: 8000 should be fine.

soninaren (commented):
I don't see an entry in https://github.com/Azure/azure-functions-templates/blob/dev/Build/PackageFiles/Dotnet_precompiled/ItemTemplates-Isolated_v4.x.nuspec. I am guessing that is on purpose.

liliankasem:
> I don't see an entry in https://github.com/Azure/azure-functions-templates/blob/dev/Build/PackageFiles/Dotnet_precompiled/ItemTemplates-Isolated_v4.x.nuspec. I am guessing that is on purpose.

I think we should probably add this, was not on purpose from my side!

Functions.Templates/Templates/McpToolTrigger-CSharp-Isolated/.template.config/vs-2017.3.host.json
soninaren: This number decides the order in which the templates show up in VS. Lower number means it is higher on the list

soninaren (approved):

fabiocav (approved):

vrdmr (approved):

PR #1716, opened 2025-11-20
Updating .NET item templates to use primary constructors

mattchenderson:
Updating the .NET templates to use primary constructors to reduce verbosity a little. The changes here require C#12, which means the .NET 8 SDK or later. If we accept these changes, we must ensure that they don't get added to older versions in the tooling feed, for example. We also need to look at some separation in the Core Tools.

This PR is being open as a draft until those enforcement mechanisms can be properly in place. At time of opening, the tooling feed generator has been updated to skip out-of-support versions. However, there are enough manual changes in that repo that we might need to add extra enforcement - I'm currently thinking of auto-review from Copilot, with an appropriate `copilot-instructions.md` to frame this new requirement and help avoid mistakes. I do not yet have a plan for the Core Tools.

### DIFF
--- a/Functions.Templates/Templates/BlobTrigger-CSharp-Isolated/BlobTriggerCSharp.cs
+++ b/Functions.Templates/Templates/BlobTrigger-CSharp-Isolated/BlobTriggerCSharp.cs
@@ -5,20 +5,13 @@ using Microsoft.Extensions.Logging;
 
 namespace Company.Function;
 
-public class BlobTriggerCSharp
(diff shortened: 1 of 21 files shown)


manvkaur:
Hi @mattchenderson , Is this PR ready for review and merge?

mattchenderson:
Closing as stale. The enforcement mechanisms described in the description are still the right next step, but I'll leave it to the team to own an equivalent PR to this one once things are properly lined up.

PR #1761, opened 2026-03-31
Replace gulp build with PowerShell script

manvkaur:
﻿ ## Summary
Fixes - https://github.com/Azure/azure-functions-bucees-planning/issues/767
 Migrate build system from gulp and npm to pure PowerShell 7.x to eliminate npm dependencies and their associated update overhead.

 ## Motivation
 - Remove npm dependency management burden
 - Eliminate need to patch npm packages
 - Simplify build process with native PowerShell

 ## Changes

 ### Added
 - Build\build-all.ps1: PowerShell 7.x build script (~650 lines) that replicates all 13 gulp tasks

 ### Removed
 - Build\gulpfile.js: Main gulp task definitions
 - Build\package.json: NPM package configuration
 - Build\package-lock.json: NPM lock file (244KB)
 - Build\gulp-utils\: Custom gulp plugins (gulp-resx-js, gulp-lcl-js)
 - package-lock.json (root): Root NPM lock file

 ### Modified
 - eng\ci\templates\jobs\build.yml: Replaced NodeTool, Npm, and Gulp tasks with single PowerShell@2 task
 - README.md: Updated build requirements from Node.js + Gulp to PowerShell 7.0+

 ## Build Verification
 - Tested PowerShell build against gulp build-all output
 - All 258 files across 6 artifacts are byte-identical
 - Handles edge cases: empty files, JSON ordering, CRLF, UTF-8 BOM, LCL XML namespaces

 ## Build Instructions
     cd Build
     .\build-all.ps1


TsuyoshiUshio (commented):
Thanks for taking this on. I ran the new PowerShell build locally, also ran the base gulp build, and the extracted extension bundle ZIP contents matched exactly (`Zip content differences=0`). Nice job preserving output parity while removing most of the gulp/package-lock dependency surface.

A couple of concerns to consider before merging:

**Concern — npm dependency is still present but now implicit**

`eng/ci/templates/jobs/build.yml` removes the explicit `NodeTool@0` / `Npm@1` setup, but `eng/ci/templates/steps/validate-schemas.yml` still runs `npm install -g ajv-cli ajv-formats` and then calls `ajv validate`.

That means the pipeline still depends on `npm`, despite the PR goal of moving to pure PowerShell to eliminate npm dependency management. It also makes the Node/npm requirement implicit on the hosted image rather than explicit in the pipeline.

Could we either convert schema validation away from `npm` too, or keep/document the Node setup until that step is migrated?

**Concern — `nuget.exe` failures in parallel blocks should be checked explicitly**

In `Build/build-all.ps1`, `nuget pack` is invoked inside `ForEach-Object -Parallel` and piped to `Out-Null`, but the parallel script blocks do not explicitly check `$LASTEXITCODE`.

`$PSNativeCommandUseErrorActionPreference` is set in the parent script, but native command behavior inside parallel runspaces is easy to get wrong. Since missing or failed template packages would be a high-impact build issue, it would be safer to check `$LASTEXITCODE` immediately after each `nuget pack` call and `throw` with the nuspec path and exit code.

TsuyoshiUshio (approved):
LGTM

PR #1793, opened 2026-05-29
manifest - update for vs code extension

manvkaur:
## Description

```
Remove/change -

- From Resource - 
     - Remove Agentframework, Arm, Bicep, Terraform
     - Mcp -> MCP
- From Use Case - Scheduled Tasks, Messaging
      - Starrters (make plural)
      - Data processing - Lets remove it as it only duplicates for Event processing
      - What even is Serverless?
      - AI and ML should be split into GenAI and Machine Learning
              - Remove Machine Learning for now.
```

### When adding or updating templates

- [ ] `.nuspec` updated with new templates
- [ ] `Resources.resx` updated with template metadata
- [ ] Or: only updating an existing template

### When updating `Functions.Templates\Template-Manifest\manifest.json`

- [ ] Schema validation passes (`Test-Json` against `manifest.schema.json`)
- [x] Ran `pwsh ./eng/scripts/validate-manifest-refs.ps1` locally and all refs resolve
- [ ] Priority tiers in `docs/priority-tiers.md` are consistent with manifest entries

### DIFF
--- a/Functions.Templates/Template-Manifest/schemas/manifest.schema.json
+++ b/Functions.Templates/Template-Manifest/schemas/manifest.schema.json
@@ -144,13 +144,9 @@
             "mysql",
             "signalr",
             "rabbitmq",
-            "mcp",
+            "MCP",
             "durable",
-            "agentframework",
-            "generic",
-            "arm",
-            "bicep",
-            "terraform"
+            "generic"
           ],
           "description": "Azure resource type the template interacts with"
         },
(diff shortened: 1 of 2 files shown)


Functions.Templates/Template-Manifest/schemas/manifest.schema.json
```diff
@@ -144,13 +144,9 @@
...
             "rabbitmq",
-            "mcp",
+            "MCP",
             "durable",
-            "agentframework",
-            "generic",
-            "arm",
-            "bicep",
-            "terraform"
+            "generic"
```
vrdmr, line 149: Lets do this - other?

vrdmr (approved):

Functions.Templates/Template-Manifest/schemas/manifest.schema.json
manvkaur: its schema, not used for now and VS code shouldn't be relying on it. Its when generic trigger or binding samples are added

